PR #87, opened 2024-01-24
When alignment fails, still returns the content

sidravi1:
Reviewer: @amiraliemami 

Estimate: 10 mins

---

## Ticket

N/A

### Goal

When alignment fails, we don't sent the LLM response but 
This was a suggestion @amiraliemami had that makes total sense! W

## How has this been tested?

Test updated and pass.

## Checklist

### DIFF
--- a/core_backend/app/llm_call/check_output.py
+++ b/core_backend/app/llm_call/check_output.py
@@ -22,11 +22,6 @@ from .utils import _ask_llm_async
 
 logger = setup_logger("OUTPUT RAILS")
 
-STANDARD_FAILURE_MESSAGE = (
-    "Sorry, I am unable to find an answer to your question in my knowledge base. "
-    "Please rephrase your question or ask a different one."
-)
-
 
 class AlignScoreData(TypedDict):
     """
@@ -104,10 +99,8 @@ async def _check_align_score(
     }
 
     if align_score.score < float(ALIGN_SCORE_THRESHOLD):
-        llm_response.llm_response = (
-            "We are unable to asnwer your question but found some content "
-            "that may be helpful"
-        )
+        llm_response.llm_response = None
+
     llm_response.debug_info["factual_consistency"] = factual_consistency.copy()
 
     return llm_response
--- a/core_backend/app/llm_call/check_output.py
+++ b/core_backend/app/llm_call/check_output.py
@@ -13,7 +13,6 @@ from ..configs.app_config import (
 )
 from ..configs.llm_prompts import AlignmentScore
 from ..schemas import (
-    ErrorType,
     UserQueryRefined,
     UserQueryResponse,
     UserQueryResponseError,
@@ -75,7 +74,7 @@ def check_align_score(func: Callable) -> Callable:
 
 async def _check_align_score(
     llm_response: UserQueryResponse,
-) -> UserQueryResponse | UserQueryResponseError:
+) -> UserQueryResponse:
     """
     Check the alignment score
     """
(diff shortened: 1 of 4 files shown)


suzinyou (approved):
Looks good. Would be nice to merge for demo as well!

core_backend/app/llm_call/check_output.py
```diff
@@ -105,17 +104,10 @@ async def _check_align_score(
...
-            error_message=STANDARD_FAILURE_MESSAGE,
-            error_type=ErrorType.ALIGNMENT_TOO_LOW,
-            debug_info=llm_response.debug_info.copy(),
-        )
-        error_response.debug_info["factual_consistency"] = factual_consistency.copy()
-        return error_response
-    else:
-        llm_response.debug_info["factual_consistency"] = factual_consistency.copy()
-        return llm_response
+        llm_response.llm_response = ""
```
amiraliemami, line 107: Here we can replace the LLM response with `STANDARD_FAILURE_MESSAGE` right? Better than blank. I would also rewrite that response to say:

    STANDARD_FAILURE_MESSAGE = (
        "Sorry, I am unable to find an answer to your question in my knowledge base."

    )

And remove the "Please rephrase your question or ask a different one." - what to do next should be left up to the chat manager.

amiraliemami: @suzinyou what do you think of this? This makes the most sense from a chat manager perspective (like our flow in Botpress)

core_backend/tests/api/test_question_answer.py
```diff
@@ -272,9 +272,9 @@ async def mock_get_align_score(*args: Any, **kwargs: Any) -> AlignmentScore:
             0.7,
         )
         update_query_response = await _check_align_score(user_query_response)
-        assert isinstance(update_query_response, UserQueryResponseError)
-        assert update_query_response.error_type == ErrorType.ALIGNMENT_TOO_LOW
+        assert isinstance(update_query_response, UserQueryResponse)
         assert update_query_response.debug_info["factual_consistency"]["score"] == 0.2
+        assert update_query_response.llm_response == ""
```
amiraliemami, line 277: See other comment re. `STANDARD_FAILURE_MESSAGE` instead of blank

amiraliemami (changes requested):
Thanks! See comments re. replacement text.

core_backend/app/llm_call/check_output.py
suzinyou: Is the idea to send only reason for error, and chat manager handles error message to send to user? If so leaving llm_response blank seems to make sense?

suzinyou: wait, why don't we return an ErrorResponse instance if the check fails?? If we send error response and error code about Alignscore check failure, chat managers and handle the user-facing message?

amiraliemami: Because we want the user to still be able to browse the FAQs even if the summarization didn't work well. For that, we need to send the `content_response` field, which we only send when we respond without erroring out. 

I think it makes sense for the alignscore to just be an internal cleaner of the LLM response. On botpress, I can also check if the LLM response is the standard "didn't find anything" response and replace that message with whatever I want while still having access to the FAQs.

sidravi1: The only problem is that we create coupling between botpress and AAQ. So every time we change the STANDARD_FAILURE_MESSAGE, we'll need to update botpress. 

Keeping it blank means, Botpress can see that no LLM response was provided and do as it sees fit. It may be even better to return None instead of "".

What are your thoughts @amiraliemami and @suzinyou?

sidravi1: And agree with @amiraliemami that sending back an ErrorResponse means it's an error. It shouldn't be an error unless we want it handled like one. Should just be like the other rails and modify the response.

amiraliemami: I see, makes sense re. the coupling. "None" would work!

amiraliemami (approved):

core_backend/app/llm_call/check_output.py
sidravi1: Ok I'll make it None.

PR #56, opened 2023-12-13
Gaurdrails tests

sidravi1:
Reviewer: @suzinyou 

Estimate: 1 hr

----

# Adding test for rails

## Ticket

Fixes: [AAQ-224](https://idinsight.atlassian.net/browse/AAQ-224)

### Goal

In the previous PR, I added rails. These add tests we can run when we change models to make sure 
the rails still run.

### Changes
- new tests
- edits to ensure it is not run automatically or in CI/CD since it calls openai and therefore costs money

## How Has This Been Tested?

* Make sure you have exported your OPENAI key
`export OPENAI_API_KEY=<YOUR_KEY>`
* Run the `rails` test
`pytest -v -m rails`

## Next up
- Docs! I'll raise another PR

## Checklist:

This checklist is a useful reminder of small things that can easily be forgotten.
Put an `x` in all the items that apply and remove any items that are not relevant to this PR.

- [x] My code follows the style guidelines of this project
- [x] I have reviewed my own code to ensure good quality
- [x] I have tested the functionality of my code to ensure it works as intended
- [x] I have resolved merge conflicts
- [x] I have updated the automated tests (if applicable)
- [x] I have written [good commit messages][1]
- [ ] I have updated the README file (if applicable)
- [ ] I have updated affected documentation (if applicable)

[1]: http://chris.beams.io/posts/git-commit/

PR #89, opened 2024-01-25
[AAQ-320] return error if llm response fails

suzinyou:
Reviewer: @amiraliemami 
Estimate:

---

## Ticket

Fixes: https://idinsight.atlassian.net/browse/AAQ-320

### Goal

If llm response generation fails, we should retrun UserQueryErrorResponse instance. Otherwise check AlignScore step gets triggered.

## To-do before merge

(Optional -- remove this section if not needed)

- [ ] Ensure PR #56 is merged

## Checklist

### DIFF
--- a/core_backend/app/configs/llm_prompts.py
+++ b/core_backend/app/configs/llm_prompts.py
@@ -122,19 +122,17 @@ PARAPHRASE_INPUT = f"""
     """
 
 # ----  Question answering bot
-
-ANSWER_QUESTION_PROMPT = """
-    You are a high-performing question answering bot.
-    You support a maternal and child health chatbot.
-
-    Answer the user query in natural language by rewording the
-    following FAQ found below. Address the question directly and do not
-    respond with anything that is outside of the context of the given FAQ.
-
-    If the FAQ doesn't seem to answer the question, respond with
-    'Sorry, no relevant information found.'
-
-    Found FAQ: {faq}"""
+ANSWER_QUESTION_ERROR_MESSAGE = "Sorry, no relevant information found."
+ANSWER_QUESTION_PROMPT = (
+    "You are a high-performing question answering bot.\n"
+    "Answer the user query in English using the information "
+    "provided in the reference text found below. "
+    "Address the question directly and do not respond with anything that is "
+    "outside of the context of the given reference text. "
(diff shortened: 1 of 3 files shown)


core_backend/app/configs/llm_prompts.py
```diff
@@ -122,19 +122,17 @@ def get_supported_languages(cls) -> list[str]:
...
-
-    Answer the user query in natural language by rewording the
-    following FAQ found below. Address the question directly and do not
-    respond with anything that is outside of the context of the given FAQ.
-
-    If the FAQ doesn't seem to answer the question, respond with
-    'Sorry, no relevant information found.'
-
-    Found FAQ: {faq}"""
+ANSWER_QUESTION_ERROR_MESSAGE = "Sorry, no relevant information found."
```
amiraliemami, line 125: We have the `STANDARD_FAILURE_MESSAGE` that we should use instead of this. We should change that to a globally accessible variable actually because it's defined twice in two different scripts. I commented on Sid's PR re. the fact that we should remove the second line from it.

core_backend/app/routers/question_answer.py
```diff
@@ -87,9 +89,17 @@ async def get_llm_answer(
...
             user_query_refined.query_text, content_response[0].retrieved_text
         )
+
+        if llm_response == ANSWER_QUESTION_ERROR_MESSAGE:
+            error_response = UserQueryResponseError(
+                query_id=response.query_id,
+                error_message=ANSWER_QUESTION_ERROR_MESSAGE,
+                error_type=ErrorType.LLM_RESPONSE_FAILED,
+            )
+            return error_response
```
amiraliemami, line 102: The reason we don't want to do this is the same as the discussion [here](https://github.com/IDinsight/aaq-core/pull/87/files#r1466206069). Even if summarisation fails, we still want to allow the user chat manager to browse the FAQs (the flow for this will be set up on the chat manager side).

amiraliemami (commented):
> If llm response generation fails, we should retrun UserQueryErrorResponse instance. Otherwise check AlignScore step gets triggered.

We're changing alignscore to not actually cause an error, so in the same sense we don't want the response to actually error out just because we couldn't summarise.

PR #125, opened 2024-03-08
Multilingual/update db

lickem22:
Reviewer: @suzinyou 

Estimate: 2h

---

## Ticket

Fixes: [AAQ-382](https://idinsight.atlassian.net/browse/AAQ-382)

### Goal
The goal of this PR is to modify the database structure and endpoints to allow  multilingual support. 
### Changes
The following changes have been made:
Changes regarding the DB models
- Added a Language model
- Updated the ContentDB model to only have the id
- Introduced ContentTextDB that contains every language version of a content.

Changes regarding the endpoints:
- Added CRUD endpoints for the LanguageModel
- Added A summary endpoints 
- Updated the Retrieve endpoint to get language version of contents
- Added an endpoint to retrieve default language

Tests have been updated and new tests have been added

## How has this been tested?
This was tested by manually testing the endpoints. New tests have been added as well. 

## Checklist

[AAQ-382]: https://idinsight.atlassian.net/browse/AAQ-382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core_backend/app/routers/manage_content.py
```diff
@@ -87,9 +113,30 @@ async def retrieve_content(
     return contents
 
 
-@router.delete("/{content_id}/delete")
+@router.get("/summary", response_model=list[ContentSummary])
```
lickem22, line 116: Endpoint to get the summary of the contents as per the frontend

core_backend/app/db/db_models.py
```diff
@@ -316,71 +321,258 @@ class ContentDB(Base):
...
+    """
+
+    __tablename__ = "languages"
+
+    language_id: Mapped[int] = mapped_column(Integer, primary_key=True, nullable=False)
+
+    language_name: Mapped[str] = mapped_column(String, nullable=False)
+    is_default: Mapped[bool] = mapped_column(Boolean, nullable=False, default=False)
+    created_datetime_utc: Mapped[datetime] = mapped_column(DateTime, nullable=False)
+    updated_datetime_utc: Mapped[datetime] = mapped_column(DateTime, nullable=False)
```
suzinyou, line 345: I learned that the advantage of mapped_column is being able to infer column types from type hints, reducing boilerplate code! (https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#using-annotated-declarative-table-type-annotated-forms-for-mapped-column )

So we don't have to import sqlalchemy specific types and can write something like
```suggestion
    language_id: Mapped[int] = mapped_column(primary_key=True)
    language_name: Mapped[str] = mapped_column()
    is_default: Mapped[bool] = mapped_column(default=False)
    created_datetime_utc: Mapped[datetime] = mapped_column()
    updated_datetime_utc: Mapped[datetime] = mapped_column()
```

if you don't use Optional[] then its by default `nullable=False` apparently..

core_backend/app/db/db_models.py
```diff
@@ -316,71 +321,258 @@ class ContentDB(Base):
...
+) -> LanguageDB:
+    """
+    Updates a language in the database
+    """
+
+    language_db = LanguageDB(
+        language_id=language_id,
+        is_default=language.is_default,
+        language_name=language.language_name,
+        updated_datetime_utc=datetime.utcnow(),
```
suzinyou, line 396: Does this not need `created_datetime_utc=language.created_datetime_utc`?

core_backend/app/db/db_models.py
suzinyou: I see all of our other code has some redundancies.. we can clean this up in another PR.

core_backend/app/db/db_models.py
```diff
@@ -316,71 +321,258 @@ class ContentDB(Base):
...
+    content_row = (await asession.execute(stmt)).scalar_one_or_none()
+    if content_row:
+        return False
+    else:
+        return True
 
 
 async def save_content_to_db(
-    content: ContentCreate,
+    content: ContentTextCreate,
```
suzinyou, line 524: Should we be explicit and name the function and variable `content_text`?

suzinyou: Should we also allow creating/deleting an entire content rather than language-by-language? cc: @Tanmay-97

core_backend/app/db/db_models.py
```diff
@@ -483,12 +768,12 @@ async def get_search_results(
...
         select(
-            ContentDB,
-            ContentDB.content_embedding.cosine_distance(question_embedding).label(
+            ContentTextDB,
+            ContentTextDB.content_embedding.cosine_distance(question_embedding).label(
                 "distance"
             ),
         )
-        .order_by(ContentDB.content_embedding.cosine_distance(question_embedding))
+        .order_by(ContentTextDB.content_embedding.cosine_distance(question_embedding))
```
suzinyou, line 776: We should first filter by language then do the vector search

core_backend/app/routers/manage_content.py
```diff
@@ -87,9 +113,29 @@ async def retrieve_content(
...
+async def retrieve_content_summary(
+    readonly_access_user: Annotated[
+        AuthenticatedUser, Depends(get_current_readonly_user)
+    ],
+    language: int,
```
suzinyou, line 121: Should we use the front-end to apply language filters? @Tanmay-97 @amiraliemami  what do you think?

core_backend/app/routers/manage_content.py
```diff
@@ -99,51 +145,121 @@ async def delete_content(
...
     record = await get_content_from_db(
-        content_id,
+        content_text_id,
         asession,
     )
 
     if not record:
         raise HTTPException(
-            status_code=404, detail=f"Content id `{content_id}` not found"
+            status_code=404, detail=f"Content id `{content_text_id}` not found"
```
suzinyou, line 154: The error code should say it's the content _text_ ID, not content ID?

core_backend/app/routers/manage_content.py
```diff
@@ -51,21 +67,31 @@ async def edit_content(
...
+            content_text_id,
+            content,
+            asession,
+        )
+
+        return _convert_record_to_schema(updated_content)
+    else:
+        raise HTTPException(
+            status_code=400,
+            detail="Content could not be updated",
```
suzinyou, line 93: Can we add the reason?

core_backend/app/routers/manage_content.py
```diff
@@ -99,51 +145,121 @@ async def delete_content(
...
     Retrieve content by id endpoint
     """
 
-    record = await get_content_from_db(content_id, asession)
+    content = await get_content_from_db(content_text_id, asession)
 
-    if not record:
+    if not content:
+        raise HTTPException(
+            status_code=404, detail=f"Content id `{content_text_id}` not found"
```
suzinyou, line 178: content text* ID

core_backend/app/routers/manage_content.py
```diff
@@ -99,51 +145,121 @@ async def delete_content(
...
+
+    if language:
+        record = await get_content_from_content_id_and_language(
+            content.content_id, int(language), asession
+        )
+        if not record:
+            raise HTTPException(
+                status_code=404,
+                detail=f"""Content `{content_text_id}`
+                with language id `{language}` not found""",
```
suzinyou, line 189: Content text* ID, but we can also add what the content ID was I suppose.

core_backend/app/routers/manage_content.py
```diff
@@ -99,51 +145,121 @@ async def delete_content(
...
+    is_updated_language: bool = True,
+) -> bool:
+    contents = await get_all_languages_version_of_content(
+        content.content_id, asession=asession
+    )
+    if len(contents) < 1:
+        if is_edit or content.content_id != 0:
+            raise HTTPException(
+                status_code=400,
+                detail=f"Content id `{content.content_id}` does not exist",
```
suzinyou, line 212: content text*

core_backend/app/routers/manage_content.py
```diff
@@ -99,51 +145,121 @@ async def delete_content(
...
+        )
+        return [_convert_record_to_schema(record) for record in records]
+
+
+async def is_content_and_language_valid(
+    content: ContentTextCreate,
+    asession: AsyncSession,
+    is_edit: bool = False,
+    is_updated_language: bool = True,
+) -> bool:
```
suzinyou, line 204: Are your pre-commit hooks running? I think interrogate should have raised an error for missing docstring. 

Also, I am finding the conditionals are a bit difficult to parse, so maybe listing out all the invalid/valid cases here would be helpful.

core_backend/app/routers/manage_content.py
```diff
@@ -51,21 +67,31 @@ async def edit_content(
...
-    updated_content = await update_content_in_db(
-        content_id,
-        content,
-        asession,
-    )
+    is_updated_language = old_content.language_id != content.language_id
 
-    return _convert_record_to_schema(updated_content)
+    if await is_content_and_language_valid(
+        content, asession, True, is_updated_language
```
suzinyou, line 81: Can we add the parameter names, e.g. what is being set as True

core_backend/app/routers/manage_content.py
```diff
@@ -99,51 +145,121 @@ async def delete_content(
...
+    content: ContentTextCreate,
+    asession: AsyncSession,
+    is_edit: bool = False,
+    is_updated_language: bool = True,
+) -> bool:
+    contents = await get_all_languages_version_of_content(
+        content.content_id, asession=asession
+    )
+    if len(contents) < 1:
+        if is_edit or content.content_id != 0:
```
suzinyou, line 209: I see why we're checking for `is_edit`: if we're trying to update a content that doesn't exist (by ID) it shouldn't work.

But what is the content_id != 0 part checking for?

suzinyou (changes requested):
Thank you so much @lickem22 ! You did so much!!

Leaving initial comments before I look at the tests or run them.

Next time we work on something this big, let's definitely break it up into smaller PRs (we can keep a feature branch that we merge to), or at least raise a draft PR.

core_backend/app/db/db_models.py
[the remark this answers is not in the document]
lickem22: Yes! that's exactly what Griffin's refactor suggestion will help us to do!

core_backend/app/db/db_models.py
```diff
@@ -316,71 +321,258 @@ class ContentDB(Base):
...
+    updated_datetime_utc: Mapped[datetime] = mapped_column(DateTime, nullable=False)
+
+    __table_args__ = (
+        Index(
+            "ix_languages_is_default_true",
+            "is_default",
+            unique=True,
+            postgresql_where=(is_default.is_(True)),
+        ),
+    )
```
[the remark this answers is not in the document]
lickem22: Yes!

core_backend/app/db/db_models.py
lickem22: Yes! I think so as well.  We can do that in the refactor PR.

core_backend/app/routers/manage_content.py
lickem22: When you say here you mean on the docstring?

core_backend/app/routers/manage_content.py
lickem22: We are adding a new content text only if content.content_id==0. If not, we add the content text to an existing content_id. So, when we are editing, we need to make sure that content.content_id!=0. But when creating, we can allow content.content_id==0.

core_backend/app/db/db_models.py
lickem22: Since we are trying to update, not adding it won't affect the DB.

core_backend/app/routers/manage_content.py
lickem22: I am resolving this for now!

core_backend/app/db/db_models.py
suzinyou: On the last point, we discussed that for now we'll keep it this way. Limiting interactions with content table seems to make sense. Can make any necessary changes later

core_backend/app/routers/manage_language.py
```diff
@@ -0,0 +1,181 @@
...
+    asession: AsyncSession = Depends(get_async_session),
+) -> LanguageRetrieve | None:
+    """
+    Create language endpoint. Upsert language to PG database
+    """
+    language.language_name = language.language_name.upper()
+    if not (await is_language_name_unique(language.language_name, asession)):
+        raise HTTPException(status_code=400, detail="Language name already exists")
+
+    if language.is_default is True:
```
suzinyou, line 41:
```suggestion
    if language.is_default:
```

suzinyou: But not necesary

core_backend/app/routers/manage_language.py
```diff
@@ -0,0 +1,181 @@
...
+        await unset_default_language(asession)
```
suzinyou, line 42: This means that currently it is possible to have no default language. I guess we can prevent that from the frontend, but not sure if we should be allowing it from the backend! What do you think?

Maybe something you can check with @thegeektets ?

If we want to always ensure at least 1 language is set to default, here we can add
```
elif not default_language_exists():
    raise HTTPException(...)
```

suzinyou: We can address in another PR when we redeisng APIs

core_backend/app/routers/manage_language.py
```diff
@@ -0,0 +1,181 @@
...
+            status_code=404, detail=f"Language id `{language_id}` not found"
+        )
+    language.language_name = language.language_name.upper()
+    if (old_language.language_name != language.language_name) and (
+        not (await is_language_name_unique(language.language_name, asession))
+    ):
+        raise HTTPException(status_code=400, detail="Language name already exists")
+
+    if language.is_default is True:
+        await unset_default_language(asession)
```
suzinyou, line 77: Here also we can edit an existing default language to be not default, resulting in only non-default languages..

core_backend/app/schemas.py
```diff
@@ -103,42 +109,78 @@ class FeedbackBase(BaseModel):
...
     """
     Pydantic model for content creation
     """
 
     # Ensure len("*{title}*\n\n{text}") <= 1600
     content_title: Annotated[str, StringConstraints(max_length=150)]
     content_text: Annotated[str, StringConstraints(max_length=2000)]
-    content_language: str = "ENGLISH"
+    language_id: int
+    content_id: Optional[int] = None
```
suzinyou, line 150: Can we rename this to `existing_content_id`?
This way it is less confusing to users. Users may supply `content_id` even for a content that doesn't exist, and receive a new `content_id` like the following:
image

suzinyou: Also some description will be helpful!
```python
from pydantic import BaseModel, Field
```

```suggestion
    existing_content_id: Optional[int] = Field(default=None, description='If adding or editing content text to an existing content, provide its ID')
```

core_backend/app/db/db_models.py
```diff
@@ -390,50 +584,143 @@ async def update_content_in_db(
 
 
 async def delete_content_from_db(
```
suzinyou, line 586: Can we rename these functions to `delete_content_text_from_db` etc? Make explicit we're dealing with content_text, not content?

core_backend/app/db/db_models.py
```diff
@@ -316,71 +321,260 @@ class ContentDB(Base):
     SQL Alchemy data model for content
     """
 
-    __tablename__ = "content"
+    __tablename__ = "contents"
 
     content_id: Mapped[int] = mapped_column(Integer, primary_key=True, nullable=False)
+
```
suzinyou, line 327: Let's add content-level metadata?

core_backend/app/routers/manage_content.py
```diff
@@ -51,21 +71,33 @@ async def edit_content(
...
-        content_id,
-        content,
-        asession,
+
+    is_valid, reason = await validate_edit_content(
+        content=content,
+        asession=asession,
+        old_content=old_content,
     )
+    if is_valid:
```
suzinyou, line 88: If I update an existing content text by associating it with a different content_id, then we can be left with a content_id in the content table that has no content text associated.

After updating, we can clean up dangling contents if they exist?

core_backend/app/routers/manage_language.py
```diff
@@ -0,0 +1,181 @@
...
+    if not language:
+        raise HTTPException(
+            status_code=404, detail=f"Language id `{language_id}` not found"
+        )
+
+    if language.is_default:
+        raise HTTPException(
+            status_code=400, detail="Default language cannot be deleted"
+        )
+    await delete_language_from_db(language_id, asession)
```
suzinyou, line 156: Here also, we should clean up any dangling contents, in case this delete operation gets rid of the last content text of a content..

suzinyou (commented):
Still reviewing but leaving first batch of comments! There are also some unresolved comments from yesterday as well. Thanks @lickem22 !!

core_backend/app/routers/manage_language.py
lickem22: I can update that to make sure there is always a default language in the DB. 
About your point, If I remember correctly there is software development [principle](https://www.highlight.io/blog/3-levels-of-data-validation-in-a-full-stack-application-with-react) stating that there should be three layers of validation:
- On the DB level,
- On the Backend level
- On the frontend level

core_backend/app/routers/manage_language.py
lickem22: Will fix that !

core_backend/app/db/db_models.py
lickem22: I'm strongly against that. Because so far we do not interact with contents but only with content_texts. Meaning that we are treating each content_texts as its own content and we only use contents to group them together. Adding contents will bring confusion as we will now have to treat them as two separate entities. Also, if we add it on contents we would have to start adding CRUD endpoints to contents as well which will bring confusion. But I am open to adding it if it is really important.

core_backend/app/routers/manage_content.py
lickem22: I will add that on update as well. Thanks for catching that.

core_backend/app/routers/manage_language.py
lickem22: I do that in the `delete_language_from_db` function

core_backend/app/routers/manage_language.py
lickem22: About the default language. I do not think it should be a requirement to always have a a default language. But We should make sure we don't have more than one. But definitely we should throw an error when we are trying to retrieve the default language but it doesn't exist in the DB.

core_backend/app/routers/manage_content.py
```diff
@@ -87,9 +120,29 @@ async def retrieve_content(
     return contents
 
 
-@router.delete("/{content_id}/delete")
+@router.get("/summary", response_model=list[ContentSummary])
+async def retrieve_content_summary(
+    readonly_access_user: Annotated[
+        AuthenticatedUser, Depends(get_current_readonly_user)
+    ],
+    language: int,
```
suzinyou, line 128: We should change the name to this!
```suggestion
    language_id: int,
```

core_backend/app/schemas.py
```diff
@@ -103,42 +110,82 @@ class FeedbackBase(BaseModel):
...
     """
-    Pydantic model for content retrieval
+    Pydantic model
+    for content summary
     """
 
+    content_text_id: int
     content_id: int
+    content_title: str
+    languages: list[str]
```
suzinyou, line 181: can we rename this to `available_languages`? Because the content summary endpoint requires a language_id, seeing languages returned was a bit confusing. Or add some JSON schema description?

core_backend/app/db/db_models.py
```diff
@@ -316,124 +321,417 @@ class ContentDB(Base):
...
+    if limit is not None:
+        stmt = stmt.limit(limit)
+    content_rows = (await asession.execute(stmt)).all()
+
+    return [c[0] for c in content_rows] if content_rows else []
+
+
+async def get_summary_of_content_from_db(
+    asession: AsyncSession,
+    language_id: int,
```
suzinyou, line 646: Right now I can  query /summary endpoint with an nonexistent language ID and would get an empty list. We should check if the language ID exists, and if not raise an error. For languages that do exist, returning empty list makes sense!

core_backend/app/routers/manage_content.py
suzinyou: Can you make this either "Content text {content_text_id}..." or "Content ID {content.content_id}..."?

core_backend/app/routers/manage_content.py
```diff
@@ -99,51 +152,146 @@ async def delete_content(
...
+
+    if language:
+        record = await get_content_from_content_id_and_language(
+            content.content_id, int(language), asession
         )
+        if not record:
+            raise HTTPException(
+                status_code=404,
+                detail=f"""Content `{content_text_id}`
+                with language id `{language}` not found""",
```
suzinyou, line 197: This error message is multilined! Something like the following should work
```
                detail= (
                    f"Content {content.content_id} "
                    f"with language id {language_id} not found."
                    f"(content text ID {content_text_id}"
                ),
```

core_backend/app/db/db_models.py
```diff
@@ -316,124 +321,417 @@ class ContentDB(Base):
...
+
+    content_stmt = select(ContentTextDB).where(ContentTextDB.content_id == content_id)
+    content_row = (await asession.execute(content_stmt)).first()
+    if not content_row:
+        stmt = delete(ContentDB).where(ContentDB.content_id == content_id)
+        await asession.execute(stmt)
     await asession.commit()
 
 
 async def get_content_from_db(
```
suzinyou, line 616: Le'ts rename to content_text_from_db

core_backend/app/routers/manage_content.py
```diff
@@ -99,51 +145,121 @@ async def delete_content(
...
+)
 async def retrieve_content_by_id(
-    content_id: int,
+    content_text_id: int,
     readonly_access_user: Annotated[
         AuthenticatedUser, Depends(get_current_readonly_user)
     ],
     asession: AsyncSession = Depends(get_async_session),
-) -> ContentRetrieve:
+    language: Optional[str] = None,
```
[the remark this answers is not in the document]
suzinyou: Should this be `language_id: Optional[int]`? Everywhere else we use `language_id`.

Also I wonder if it's worth using ISO codes as language IDs. This way users can easily map language ID to language name..

core_backend/app/db/db_models.py
suzinyou: We will be adding content-level information like tags down the line, but we can add it when we get to it!

core_backend/app/routers/manage_language.py
suzinyou: Don't we need that information for the front-end? @Tanmay-97

core_backend/app/db/db_models.py
suzinyou: Btw feel free to turn some of these into backlog tickets!

core_backend/app/routers/manage_content.py
```diff
@@ -99,51 +152,146 @@ async def delete_content(
...
         )
-    await delete_content_from_db(content_id, asession)
+
+    await delete_content_text_from_db(content_text_id, record.content_id, asession)
 
 
-@router.get("/{content_id}", response_model=ContentRetrieve)
+@router.get(
+    "/{content_text_id}", response_model=Union[ContentRetrieve, List[ContentRetrieve]]
+)
```
suzinyou, line 169: Given our discussion on slack, should we just remove this endpoint for now?

core_backend/app/routers/manage_content.py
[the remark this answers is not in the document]
lickem22: I am wondering if it is best practice. Will do more research on that

PR #132, opened 2024-03-18
Multilingual db update

lickem22:
Reviewer: @amiraliemami  

Estimate: 1h

---

## Ticket

Fixes: [AAQ-382](https://idinsight.atlassian.net/browse/AAQ-382)

### Goal
This PR is the first official PR as part of the multilingual support. The goal is to add the language table as well as the routes to be able to run CRUD operations. 
### Changes
Added a  language submodule with the following files:
- schemas
- routers
- models
Added tests for languages
Also, since we are not planning on building a frontend for the languages yet, added english as the default language in the migration script.
## How has this been tested?
This has been tested by manually testing the endpoints one by one and updating the tests.
## Checklist

- [] I have updated the README file (if applicable)

[AAQ-382]: https://idinsight.atlassian.net/browse/AAQ-382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core_backend/app/languages/routers.py
```diff
@@ -0,0 +1,182 @@
...
+    save_language_to_db,
+    update_language_in_db,
+)
+from .schemas import LanguageBase, LanguageRetrieve
+
+router = APIRouter(prefix="/languages")
+logger = setup_logger()
+
+
+@router.post("/", response_model=LanguageRetrieve)
```
lickem22, line 27: Made changes for the endpoint as per [this thread](https://idinsight.slack.com/archives/C05RSDLN1GA/p1710490893253359)

core_backend/migrations/versions/1ff08438751b_add_language_table.py
```diff
@@ -0,0 +1,59 @@
...
+    op.create_index(
+        "ix_languages_is_default_true",
+        "languages",
+        ["is_default"],
+        unique=True,
+        postgresql_where=sa.text("is_default IS true"),
+    )
+    # add default language
+    date = str(datetime.now())
+    op.execute(
```
lickem22, line 43: Added this script to add english as a defaut language. This is because we want to separate the language frontend from this feature since we haven't created wireframes yet.

core_backend/app/languages/routers.py
```diff
@@ -0,0 +1,182 @@
...
+) -> LanguageRetrieve | None:
+    """
+    Create language endpoint. Upsert language to PG database
+    """
+    language.language_name = language.language_name.upper()
+    if not (await is_language_name_unique(language.language_name, asession)):
+        raise HTTPException(status_code=400, detail="Language name already exists")
+
+    if language.is_default is True:
+        await unset_default_language(asession)
```
amiraliemami, line 43: Nice one!

core_backend/app/languages/routers.py
```diff
@@ -0,0 +1,182 @@
...
+    language: LanguageBase,
+    full_access_user: Annotated[
+        AuthenticatedUser, Depends(get_current_fullaccess_user)
+    ],
+    asession: AsyncSession = Depends(get_async_session),
+) -> LanguageRetrieve | None:
+    """
+    Create language endpoint. Upsert language to PG database
+    """
+    language.language_name = language.language_name.upper()
```
amiraliemami, line 38: We should document the fact that we force language names into uppercase

core_backend/app/languages/routers.py
```diff
@@ -0,0 +1,182 @@
...
+    if language.is_default:
+        await unset_default_language(asession)
+
+    updated_language = await update_language_in_db(
+        language_id,
+        language,
+        asession,
+    )
+
+    return _convert_language_record_to_schema(updated_language)
```
amiraliemami, line 86: Two comments here:

1. Should we split this up into two endpoints? One for changing the language name and one for setting as default. This might make more sense given the use-cases.
2. I think we should not give the option to set is_default to false because that means we are left with no language being default in the database. You should only be able to set a given language to default, since the app logic handles unsetting the previously default language.

amiraliemami (changes requested):

core_backend/app/languages/routers.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    Create language endpoint. Upsert language to PG database
+    """
+    language.language_name = language.language_name.upper()
+    if not (await is_language_name_unique(language.language_name, asession)):
+        raise HTTPException(status_code=400, detail="Language name already exists")
+
+    if language.is_default is True:
+        await unset_default_language(asession)
+
```
amiraliemami, line 44: Maybe we should have another check that checks if there is a default language and if not, returns an error and says "No default language found. Please either create a default language or set an existing language as default."? Not sure about this pattern though. Any other ideas for how to ensure there is always a default language?

amiraliemami (dismissed):
.

amiraliemami (commented):
Thanks for this - really nice code and logic and tests! 

Two points re. default language:
- Check out the comment on the "edit_language" router - I think that should be fixed to avoid causing a situation where there is no default language. 
- Maybe we also add some logic onto the "create_language" endpoint to ensure there is always a default language?

core_backend/app/languages/routers.py
lickem22: Yes. Maybe in blogpost or later when we will finalize multilingual support

core_backend/app/languages/routers.py
lickem22: Right now  I can just make sure we cannot update the default language to non_default. Telling the user to directly set the new default language.

core_backend/app/languages/routers.py
lickem22: I think it is okay for now. But I am fine with doing more thinking and updates when we will actually build a frontend for languages management. Then we will have a clear understanding of what to do.

lickem22:
> Thanks for this - really nice code and logic and tests!
> 
> Two points re. default language:
> 
> * Check out the comment on the "edit_language" router - I think that should be fixed to avoid causing a situation where there is no default language.
> * Maybe we also add some logic onto the "create_language" endpoint to ensure there is always a default language?

About checking for a default language, we are already adding a default language in the migrations, so it shouldn't be an issue. So I suggest we keep it like this now and when we will build the language for frontend we will add the logic, by for example creating the language at launch if there is no language in the DB.

amiraliemami (approved):
Thanks for the changes!

PR #145, opened 2024-04-04
Alignscore: better logs and fix for "state" field

sidravi1:
Reviewer: @suzinyou 
Estimate: 10 mins

---

## Ticket

Fixes: [AAQ-369](https://idinsight.atlassian.net/browse/AAQ-369)

## Description

Better logs when alignscore fails.
Also fixes the state field error

### Changes
- update to tests
- update to logging

## How has this been tested?
- test cases pass

## Checklist

[AAQ-369]: https://idinsight.atlassian.net/browse/AAQ-369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core_backend/app/llm_call/check_output.py
+++ b/core_backend/app/llm_call/check_output.py
@@ -79,7 +79,7 @@ async def _check_align_score(
     evidence = _build_evidence(llm_response)
     claim = llm_response.llm_response
     assert claim is not None, "LLM response is None"
-    align_score_date = AlignScoreData(evidence=evidence, claim=claim)
+    align_score_data = AlignScoreData(evidence=evidence, claim=claim)
 
     if ALIGN_SCORE_METHOD is None:
         logger.warning(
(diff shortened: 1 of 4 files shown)


core_backend/app/llm_call/check_output.py
```diff
@@ -146,7 +160,7 @@ async def _get_llm_align_score(align_score_data: AlignScoreData) -> AlignmentSco
     try:
         alignment_score = AlignmentScore.model_validate_json(result)
     except ValidationError as e:
-        logger.error(f"LLM alignment score respone is not valid json: {e}")
+        logger.error(f"LLM alignment score response is not valid json: {e}")
 
     logger.info(f"LLM Alignment result: {alignment_score.model_dump_json()}")
```
suzinyou, line 165: BTW, this line just repeats the llm output log, should we delete it?
image

core_backend/app/llm_call/check_output.py
```diff
@@ -105,8 +109,18 @@ async def _check_align_score(
     }
 
     if align_score.score < float(ALIGN_SCORE_THRESHOLD):
```
suzinyou, line 111: I am having trouble coming up with a test case for this..  RAG needs to succeed, but alignment needs to fail!

suzinyou (approved):
Approved.. But I could not test all logic :(

core_backend/app/llm_call/check_output.py
sidravi1: I think the first one is from lite_llm logging and we may turn that off at some stage. And the second one is from us. 
My vote is to let the both stay.

core_backend/app/llm_call/check_output.py
sidravi1: What if we just call `_check_align_score` with a made up `UserQueryResponse`?
It more a unit test than an end to end test.

PR #173, opened 2024-04-26
Multi-user content part 2: Add user table and generate user_ids

amiraliemami:
Reviewer: @sidravi1 
Estimate: 1hr

---

## Ticket

Fixes: Completes [476](https://idinsight.atlassian.net/browse/AAQ-476?atlOrigin=eyJpIjoiY2Q4OTQxMTI4NGFmNDczNmE0YTU2Njg1NTM4MThjM2YiLCJwIjoiaiJ9)

### Goal

- Add user table that keeps unique user_ids per user. Generate user_ids per user instead of using their username as the id.

Note: Retrieval endpoints are hardcoded to use USER1's contents. Per-user retrieval will be implemented in next PR.

### Changes

- Added user table with columns for user_id and username
- content: Updated backend logic to look up user_id based on username and filter content db
- Added initialisation script `core_backend/add_users_to_db.py` to use for filling the user DB on application start (since we don't have user creation flow right now). This script is now used in:
  - `setup-dev` Make target for dev
  - core_backend Dockerfile (for local docker-compose deployment and aws)
- Also added fixture to add 1 user to the db in pytests before tests are run.
- WhatsApp endpoint / tests are disabled since there is no way to pass user information there as of now.

### Future Tasks (optional)

- Add tests for new functionality
- Mention functionality in docs
- Add unique retrieval tokens per user (next PR)

## How has this been tested?

(NOTE: remove old containers/images/volumes before testing)
- Run dev setup and test functionality
- Run docker-compose and test functionality (take fresh copy of `template.env` and update the OpenAI token)

## To-do before merge

- Merge #170 

## Checklist


core_backend/app/users/__init__.py
amiraliemami, line 1: This stuff definitely shouldn't be here - where do you suggest I put this initial user loading script?

core_backend/tests/api/conftest.py
amiraliemami, line 1: This is too intertwined with the app configs right now - will separate it such that pytest adds its own user and contents later. Need to move that stuff out of `__init__.py` though probably.

core_backend/tests/api/conftest.py
```diff
@@ -198,17 +199,16 @@ def fullaccess_token() -> str:
...
 @pytest.fixture(scope="session")
 def readonly_token() -> str:
     """
     Returns a token with readonly access
     """
-    return create_access_token(
-        "user1"
-    )  # TEMPORARY HARDCODED USER ID - this is also fullaccess!
+    # NOTE - this is also fullaccess!
+    return create_access_token(USER1_USERNAME)
```
amiraliemami, line 212: Not sure about this - how does it know to return a readonly token?

core_backend/migrations/versions/3cdc029625e5_added_user_id_to_contentdb.py
```diff
@@ -0,0 +1,44 @@
+"""Added user_id to ContentDB
```
sidravi1, line 1: After I'm done making all the changes and creating the migrations, I consolidate them so you just have the 1.
This is as easy as deleting all the new migrations, take a fresh db, upgrade to head, and ask it to create a new migration again. You'll have one file with all the changes.

core_backend/add_users_to_db.py
```diff
@@ -0,0 +1,43 @@
+from datetime import datetime
```
sidravi1, line 1: Is this just for testing?

core_backend/app/auth/config.py
```diff
@@ -1,10 +1,12 @@
 import os
 
-CONTENT_READONLY_PASSWORD = os.environ.get("CONTENT_READONLY_PASSWORD", "readonly")
+USER1_USERNAME = os.environ.get("USER1_USERNAME", "user1")
```
sidravi1, line 3: Why do we need this anymore? I assume this is now in the Db right?

core_backend/app/users/models.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class UserDB(Base):
+    """
+    SQL Alchemy data model for users
+    """
+
+    __tablename__ = "user"
+
+    user_id: Mapped[str] = mapped_column(String, primary_key=True, nullable=False)
```
sidravi1, line 25: Noticed no `access_level`. I'm cool with it. 
Though would we want a super_user that has access to content?

core_backend/app/users/models.py
```diff
@@ -0,0 +1,88 @@
...
+) -> UserDB:
+    """
+    Saves a user in the database
+    """
+
+    # Check if user with same username already exists
+    stmt = select(UserDB).where(UserDB.username == user.username)
+    result = await asession.execute(stmt)
+    try:
+        result.scalar_one()
```
sidravi1, line 50: /nit should this be just `.one()`. Are we adding overhead by asking it to be `scalar` or does it make more efficient?

core_backend/app/users/models.py
```diff
@@ -0,0 +1,88 @@
...
+
+class UserDB(Base):
+    """
+    SQL Alchemy data model for users
+    """
+
+    __tablename__ = "user"
+
+    user_id: Mapped[str] = mapped_column(String, primary_key=True, nullable=False)
+    username: Mapped[str] = mapped_column(String, nullable=False)
```
sidravi1, line 26: Should there be a password here? Or sorted by google oauth?

core_backend/app/users/models.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # retrieval_token: Mapped[str] = mapped_column(String, nullable=False)
+
+    created_datetime_utc: Mapped[datetime] = mapped_column(DateTime, nullable=False)
+    updated_datetime_utc: Mapped[datetime] = mapped_column(DateTime, nullable=False)
```
sidravi1, line 31: What would an update look like?

sidravi1: Suggest scrapping `updated_datetime_utc` for now if there is no use

sidravi1: Or maybe we want `last_active` or something similar

core_backend/app/contents/routers.py
```diff
@@ -35,7 +36,15 @@ async def create_content(
     upserts it to PG database
```
sidravi1, line 36: Just noticed that it says "upserts" but it just "inserts" right?

core_backend/app/users/models.py
sidravi1, line 26: What is the value in having a userid and a username? Looks like we are allowing username to be not unique and userid to unique. But later we retrieve userid by username.

core_backend/app/users/models.py
```diff
@@ -0,0 +1,88 @@
...
+async def get_user_by_username(
+    username: str,
+    asession: AsyncSession,
+) -> Optional[UserDB]:
+    """
+    Retrieves a user by username
+    """
+
+    # Check if user with same username already exists
+    stmt = select(UserDB).where(UserDB.username == username)
```
sidravi1, line 79: or is multiple users have the same username

core_backend/app/users/models.py
```diff
@@ -0,0 +1,88 @@
...
+    result = await asession.execute(stmt)
+    try:
+        user = result.scalar_one()
+        return user
+    except NoResultFound as err:
+        raise ValueError(f"User with username {username} does not exist.") from err
```
sidravi1, line 85: or if multiple users have the same username

core_backend/app/users/schemas.py
```diff
@@ -0,0 +1,17 @@
...
+class UserCreate(BaseModel):
+    """
+    Pydantic model for user creation
+    """
+
+    username: str
+    user_id: str
+    # retrieval_token: str
+    created_datetime_utc: datetime
+    updated_datetime_utc: datetime
```
sidravi1, line 15: Same here. Probably scrap this field.

core_backend/startup.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 python -m alembic upgrade head
+python add_users_to_db.py
```
sidravi1, line 3: I guess this is just in lieu of google oauth. There is no other way to create users right now?

core_backend/tests/api/test_manage_content.py
```diff
@@ -96,7 +96,7 @@ def test_edit_and_retrieve_content(
         content_title: str,
         content_text: str,
         fullaccess_token: str,
-        readonly_token: str,
+        # readonly_token: str,
```
sidravi1, line 99: Same comment as other PR. Let's think of test cases since this is getting complex.

sidravi1 (approved):
Thanks @amiraliemami!

I'm going to approve it since it not into main. Im confused by the role of `userid` and `username`. I would expect `userid` to be used everywhere including QA endpoints. Is this just a temporary thing that gets fixed in future PRs?

core_backend/migrations/versions/3cdc029625e5_added_user_id_to_contentdb.py
amiraliemami: Yeah this makes sense - I'll do this once all PRs are merged and table structures finalised before merging into main. If I do it now it might create rebasing hell.

core_backend/add_users_to_db.py
amiraliemami: This is for getting users into the user table since in the current setup we don't have user creation endpoints. This just loads the environment variables directly into the user table right when the app initialises.

I don't think it makes sense right now to create complicated user signup/addition workflows - I suspect that we can offload all of that to another service (like [Fief](https://docs.fief.dev/)) and build whatever we need to interface with that user management service when we get there (probably next sprint).

core_backend/app/auth/config.py
amiraliemami: Yeah this is in the DB, because of the `core_backend/add_users_to_db.py` script. It didn't occur to me to update this to fetch from the db instead of the env var since it's the same thing right now. 

I'm going to have to read the retrieval key from the DB in PR #179 anyway so will also read the username from there then. We can decide if we want to also store passwords in the DB at that point (I don't think so though - see point about offloading user management to another service like [Fief](https://docs.fief.dev/) above.)

core_backend/app/users/models.py
amiraliemami: I think at some point we'll want access_level and super_user, but I'm trying to keep these PRs lean. Will make a ticket for adding a super_user (since we'll definitely need that at some stage).

core_backend/app/users/models.py
amiraliemami: This was a copilot suggestion. [.scalar_one()](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Result.scalar_one) does two operations - we just need [.one()](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Result.one),  thanks for picking this up. I'll also add an exception for [MultipleResultsFound](https://docs.sqlalchemy.org/en/14/core/exceptions.html#sqlalchemy.exc.MultipleResultsFound).

core_backend/app/users/models.py
amiraliemami: See previous comment. Hoping to defer passwords to OAuth.

core_backend/app/users/models.py
amiraliemami: I added "last_active" as a todo for later - it might come in handy. For now I haven't implemented any updating mechanisms, but PR #179 adds an endpoint for updating the retrieval_key (which will be a column in this table) and then we use the `updated_datetime_utc` column.

core_backend/app/contents/routers.py
amiraliemami: Yeah, will fix.

core_backend/app/users/models.py
amiraliemami: Done

core_backend/app/users/models.py
amiraliemami: Done

core_backend/app/users/models.py
amiraliemami: Good point - in the application logic I'm assuming username is unique for now. I'm anticipating that when we offload the user management to a OAuth service, it will just send back a unique ID per user (which would replace username) and thought we might want to keep our own generated user_id?

Also, we can control and standardise user_ids but usernames can be anything... It's definitely best practice to keep a generated ID but I get that right now we essentially have 2 equivalent ID columns...

Not sure - maybe I could just use username as ID for now. Or if you're okay with it, we keep this until we get to find out how OAuth really works and adjust there.

core_backend/app/users/models.py
amiraliemami: Done

core_backend/app/users/models.py
amiraliemami: Done

core_backend/app/users/schemas.py
amiraliemami: Removed in PR #179. This should always be handled by the DB-insert function anyway.

core_backend/startup.sh
amiraliemami: Yup until we have OAuth. No other way to add users right now.

amiraliemami:
> Thanks @amiraliemami!
> 
> I'm going to approve it since it not into main. Im confused by the role of `userid` and `username`. I would expect `userid` to be used everywhere including QA endpoints. Is this just a temporary thing that gets fixed in future PRs?

- Replied in another comment about user_id vs username. Your concern/confusion is valid they are both expected to be unique. Happy to leave it like this until we find out how OAuth will interact with our app?

- `user_id` will be used everywhere, including the QA endpoints. In PR #176, I use a function `get_user_by_token()` to find the user_id I should filter on when running retrieval. See [this line](https://github.com/IDinsight/aaq-core/blob/2e0278d13a0e170c4704b4e780800041aadd77cf/core_backend/app/question_answer/routers.py#L180).

PR #175, opened 2024-04-27
Multi-user content part 2.5: Renaming and cleanup of DB tables

amiraliemami:
Reviewer: @sidravi1 
Estimate: 15mins

The table naming was getting confusing with the new "user" table and the old "user-queries..." etc. 

- This cleans that up and makes previous work also make more sense going forward.
- Also standardises up old migration commit messages and filenames (no functionality change there).

Before:
image

After:
image

## To-do before merge

- Merge #173


sidravi1 (approved):
I like it! Thanks for cleaning this up. 
Probably should also update the schemas to match the new names.

I also worry about the merge conflicts with other PRs that touch the db (like mine!) but we'll sort that out when we get to it.

amiraliemami:
> I like it! Thanks for cleaning this up. Probably should also update the schemas to match the new names.
> 
> I also worry about the merge conflicts with other PRs that touch the db (like mine!) but we'll sort that out when we get to it.

Thanks! I missed the schemas. Done.

I worry about the merge conflicts too, but have to pay the price someday. Earlier the better I guess lol

PR #176, opened 2024-04-27
Multi-user retrieval part 1: Add per-user keys and filter content before responding

amiraliemami:
Reviewer: @sidravi1 
Estimate: 30mins

---

## Ticket

Fixes: [477](https://idinsight.atlassian.net/browse/AAQ-477?atlOrigin=eyJpIjoiY2NmZTlkNWU1OTkzNDBiODkyMDdhYWYyOTA1YjhhOGMiLCJwIjoiaiJ9)

### Goal

- Give each user a unique API key which is used to only use their contents when fetching contents.

### Changes

- Give each user a retrieval key (new column in user table)
- Filter contents in QA endpoints
- Check for user_id in feedback endpoints. 
  - Q: **Should we add a user_id column to the feedback tables? Probably not, right?**
- Update deployment
- Fix pytests
- Add user_id to query table now that we know who the requests are associated with
  - **Should we add a user_id column to the query-response and error table? Probably not, right?**

### Future Tasks

- SECURITY: Hash the API keys instead of saving them in the DB as strings!
- Make new tests

## How has this been tested?

- run tests
- run dev setup and check functionality
- run deployment (with fresh `template.env` copy and removed volumes and images) and check

## To-do before merge

- Merge #175 

## Checklist


core_backend/app/auth/dependencies.py
```diff
@@ -24,24 +27,30 @@
...
         "access_level": "fullaccess",
     },
-    "readonly": {"password": CONTENT_READONLY_PASSWORD, "access_level": "readonly"},
 }
 
 
 def auth_bearer_token(
     credentials: HTTPAuthorizationCredentials = Depends(bearer),
-) -> None:
+) -> str:
```
amiraliemami, line 43: Is it reasonable for this to return the token as string? This worked so I didn't look into other ways of getting the token value into the hands of the QA/Feedback endpoints to use to fetch the user_id.

sidravi1: String should be fine I think. 

Also, should we just check for user here? In the future, you would extract token, lookup user and return that to the function right?

core_backend/app/question_answer/routers.py
```diff
@@ -53,23 +55,28 @@
...
+    asession: AsyncSession = Depends(get_async_session),
+    auth_token: str = Depends(auth_bearer_token),
 ) -> QueryResponse | JSONResponse:
     """
     LLM response creates a custom response to the question using LLM chat and the
     most similar embeddings to the user query in the vector db.
     """
-    # hardcode USER1_USERNAME here. Later will change this to "get_user_by_token()"
-    user = await get_user_by_username(USER1_USERNAME, asession)
+    user = await get_user_by_token(auth_token, asession)
```
sidravi1, line 66: Same comment as above. Ideally we don't pass token but the user. So we move this line to get_user_from_token function. Then the endpoint dont have to know or worry about tokens.

core_backend/app/question_answer/routers.py
```diff
@@ -187,54 +236,68 @@ async def get_semantic_matches(
...
     <BR>
     <B>Note</B>: If you wish to only provide `feedback_text`, don't include
     `feedback_sentiment` in the payload.
     """
+    user = await get_user_by_token(auth_token, asession)
+    if user is None:
+        return JSONResponse(
+            status_code=400,
+            content={"message": "No user associated with the provided token."},
+        )
```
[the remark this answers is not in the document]
sidravi1: Hmmm... i agree that it is not needed since feedback secret key is associated with a unique query. So if you leak your feedback secret key, someone else can provide feedback for your query.

I can't decide if that's a feature or a bug but leaning towards it's a good thing.

There is something nice about passing on a feedback key to some other service without sharing the bearer token to allow them to forward feedback.

core_backend/app/question_answer/routers.py
```diff
@@ -292,7 +322,7 @@ async def content_feedback(
                 },
             )
         await update_votes_in_db(
-            user_id=USER1_USERNAME,  # TEMPORARY HARDCODED USER ID
+            user_id=user.user_id,
```
sidravi1, line 325: Ooh. I guess you need it for content_feedback. Do you think it's worth adding to `response_feedback` for consistency?

core_backend/app/users/models.py
```diff
@@ -59,6 +59,7 @@ async def save_user_to_db(
     content_db = UserDB(
         user_id=user.user_id,
         username=user.username,
+        retrieval_key=user.retrieval_key,
```
sidravi1, line 62: At some stage we'll need a "generate_new_retrieval_key"

core_backend/app/auth/config.py
```diff
@@ -1,11 +1,14 @@
 import os
 
+# user 1
 USER1_USERNAME = os.environ.get("USER1_USERNAME", "user1")
 USER1_PASSWORD = os.environ.get("USER1_PASSWORD", "fullaccess")
+USER1_RETRIEVAL_KEY = os.environ.get("USER1_RETRIEVAL_KEY", "user1-key")
```
sidravi1, line 6: Should these be in the Db now? And we just get from there?

core_backend/migrations/versions/e2fece37020f_add_retrieval_key_column_to_user_table.py
```diff
@@ -0,0 +1,30 @@
...
+# revision identifiers, used by Alembic.
+revision: str = "e2fece37020f"
+down_revision: Union[str, None] = "55cf4d21158b"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("user", sa.Column("retrieval_key", sa.String(), nullable=False))
```
sidravi1, line 23: Maybe it's future PR thing but maybe we should keep it encrypted/hashed in the Db? And compare the hashed values.

sidravi1 (approved):
Looks good @amiraliemami!

Some minor comments. Also, I'm having trouble keeping what's transitory and whats here to stay in mind.  E.g. should all user stuff be in the db and no longer in config?

core_backend/app/question_answer/routers.py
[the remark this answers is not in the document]
amiraliemami: Hmm but to call the feedback endpoints you need the API key anyway so it's not that another service could add feedback without having some user's API key.

I think I'll leave it in for consistency with the other endpoints for now. I think it actually makes sense to check if the user exists during auth, so this will become outdated in the last PR anyway.

core_backend/app/question_answer/routers.py
amiraliemami: Yup, perfect. Will add this to final PR once auth reads from DB.

core_backend/app/question_answer/routers.py
amiraliemami: Adding the "user_id" column you mean?

core_backend/app/auth/config.py
amiraliemami: Yeah, will add this to final PR once auth reads from DB.

core_backend/app/question_answer/routers.py
amiraliemami: Keeping a note of this.

PR #179, opened 2024-04-28
Multi-user retrieval part 3: Auth from DB and endpoint for rotating API keys

amiraliemami:
Now Functional!

Reviewer: @sidravi1 
Estimate: 1hr

---

## Ticket

Fixes: [511](https://idinsight.atlassian.net/browse/AAQ-511),  [490](https://idinsight.atlassian.net/browse/AAQ-490?atlOrigin=eyJpIjoiZTllYzMxMzc5M2ZmNDA4ZGIxYTI0ZmFjZTIxNGY0OTAiLCJwIjoiaiJ9), and [492](https://idinsight.atlassian.net/browse/AAQ-492).

### Goal

1. Do auth based on user information in the DB instead of environment variables
2. Add endpoint to allow an authenticated user (with username + password) to change their retrieval key

### Changes

- Added endpoint as above
- Update authentication flow to read from the DB instead of env vars
- Added a temporary hashed_passwords column (won't be storing passwords once we move to OAuth)
- Removed readonly functionality and tests (left it in admin app incase we want to use it again)
- Some minor frontend cleanup and backend route tagging cleanup

### Future Tasks

- Add tests

## How has this been tested?

- Previous tests run
- Dev and deployment work as expected. Call this endpoint and check the database and try the new token in playground.
- Docker-compose deployment works as expected 

## Checklist


core_backend/app/users/models.py
```diff
@@ -74,6 +74,28 @@ async def save_user_to_db(
...
+    user_db: UserDB,
+    new_retrieval_key: str,
+    asession: AsyncSession,
+) -> UserDB:
+    """
+    Updates a user's API key
+    """
+
+    # test if new_retrieval_key is a 32-character string
+    if len(new_retrieval_key) != 32:
```
sidravi1, line 87: Any reason it would not be? We generate it right and not the user?

core_backend/app/users/routers.py
```diff
@@ -0,0 +1,54 @@
...
+
+    new_retrieval_key = generate_key()
+
+    try:
+        await update_user_retrieval_key(
+            user_db=user_db,
+            new_retrieval_key=new_retrieval_key,
+            asession=asession,
+        )
+    except Exception as e:
```
sidravi1, line 44: What kind of exception are we catching? The ValueError? Then we should catch just that.
If not, probably ok to crash loudly?

core_backend/app/utils.py
```diff
@@ -1,12 +1,22 @@
...
 from .config import LOG_LEVEL
 
 
+def generate_key() -> str:
+    """
+    Generate a new 32-character API key
+    """
+
+    # 24 bytes gives 32 characters here
+    return secrets.token_urlsafe(24)
```
sidravi1, line 17: Should we use 32 bytes?

https://docs.python.org/3/library/secrets.html#how-many-bytes-should-tokens-use

sidravi1 (approved):
Thanks @amiraliemami

Every PR I think of something and you have thought of it already and added it to the next PR :) Nice work mate!

sidravi1:
lol. Just realised this is in draft! Happy to quickly review it again

core_backend/app/utils.py
amiraliemami: Yeah that makes sense given the link. 32 bytes makes 43 characters which was an ugly number but its ok haha

core_backend/app/users/models.py
amiraliemami: You're right, this is probably redundant. Will remove

core_backend/app/database.py
```diff
@@ -70,3 +70,10 @@ async def get_async_session() -> AsyncGenerator[AsyncSession, None]:
         get_sqlalchemy_async_engine(), expire_on_commit=False
     ) as async_session:
         yield async_session
+
+
+async def get_async_session_direct() -> AsyncSession:
+    """Return a SQLAlchemy async session."""
+    engine = get_sqlalchemy_async_engine()
+    async_session = AsyncSession(engine, expire_on_commit=False)
+    return async_session
```
amiraliemami, line 79: Any ideas of how to clean this up somehow? Why do we use generators anyway? Was having a lot of trouble with the generators so made this new function to use with auth. Please suggest better approaches!

core_backend/app/auth/dependencies.py
```diff
@@ -9,62 +9,56 @@
...
-        return token  # NOT SURE ABOUT THIS - IS THIS REASONABLE?
 
+    asession = await get_async_session_direct()
+    try:
+        user_db = await get_user_by_token(token, asession)
+        return user_db
+    except Exception as err:
+        raise HTTPException(status_code=401, detail="Invalid bearer token") from err
+    finally:
+        await asession.aclose()
```
amiraliemami, line 40: This process of opening a session, doing something, then closing it makes sense here right?

core_backend/app/auth/config.py
```diff
@@ -1,15 +1,5 @@
 import os
 
-# user 1
```
amiraliemami, line 3: I moved all of this to the main `app.config.py` so it can be read by the user-loading script (which is a stand-in for until we have OAuth-aware user creation) and pytest, but Auth doesnt need to know about this anymore since it reads from the DB

core_backend/app/auth/dependencies.py
```diff
@@ -9,62 +9,56 @@
...
-        "access_level": "fullaccess",
-    },
-}
-
-
-def auth_bearer_token(
+async def auth_bearer_token(
     credentials: HTTPAuthorizationCredentials = Depends(bearer),
-) -> str:
+) -> UserDB:
```
amiraliemami, line 26: Now returns the DB as we discussed, and so each endpoint doesnt have to its own check. If auth succeeds, it means that the user exists and we're good to go.

core_backend/app/users/models.py
```diff
@@ -25,6 +24,7 @@ class UserDB(Base):
 
     user_id: Mapped[str] = mapped_column(String, primary_key=True, nullable=False)
     username: Mapped[str] = mapped_column(String, nullable=False)
+    hashed_password: Mapped[str] = mapped_column(String(64), nullable=False)
```
amiraliemami, line 27: Added hashed_passwords column temporarily - once we have OAuth we won't be storing passwords.

core_backend/app/database.py
sidravi1: Using the context manager makes sure that the session is always closed when done. The level where it is created is the level where it should be closed - that is best practice. You should not depend on the function calling it to clean up properly.

But in order to do that, we need to return control to the context manager. And you need a generator to do that.

I'll take a look where it is being used and see if I can suggest other ways to do it.

core_backend/app/auth/dependencies.py
```diff
@@ -9,62 +9,56 @@
...
     the question-answering endpoints
     """
     token = credentials.credentials
-    # change to "hash check against db" logic later
-    if token not in [USER1_RETRIEVAL_KEY, USER2_RETRIEVAL_KEY]:
-        raise HTTPException(status_code=401, detail="Invalid bearer token")
-    else:
-        return token  # NOT SURE ABOUT THIS - IS THIS REASONABLE?
 
+    asession = await get_async_session_direct()
```
sidravi1, line 33: Ah! you are using it here. You could also just do what I do in [this fixture](https://github.com/IDinsight/aaq-core/blob/d8f02881bb7f610336ea1bd4bda505cce0fe2642/core_backend/tests/api/test_urgency_detect.py#L43)

and use `async for`

core_backend/app/auth/dependencies.py
sidravi1: My only qualm is that session is not actually closed where it is created. Leads to weird behaviour that is hard to debug. See comment above on one way to do this cleanly.

core_backend/app/auth/dependencies.py
```diff
@@ -9,62 +9,56 @@
...
+    *, username: str, password: str
+) -> Optional[AuthenticatedUser]:
     """
     Authenticate user using username and password.
     """
-    if username not in USERS:
+    asession = await get_async_session_direct()
+    try:
+        user_db = await get_user_by_username(username, asession)
+        await asession.aclose()
```
sidravi1, line 52: oof. Yeah - not nice to have to remember this everywhere. I bet we will stuff this up at some stage and suddenly run out of session pools at scale.

core_backend/app/auth/dependencies.py
```diff
@@ -9,62 +9,56 @@
...
+        if user_db.hashed_password == get_key_hash(password):
+            # hardcode "fullaccess" now, but may use it in the future
+            return AuthenticatedUser(username=username, access_level="fullaccess")
+        else:
+            return None
+    except Exception:
```
sidravi1, line 58: Don't catch base exceptions! It leads to weird bugs where things go wrong but errors get absorbed by lines like this. Also suggest adding logging.

sidravi1, line 58: Here's an example of a place where we forgot to close the session!

core_backend/app/auth/dependencies.py
```diff
@@ -99,36 +91,17 @@ def get_current_user(
...
-            username=username,
-            access_level=cast(AccessLevel, USERS[username]["access_level"]),
-        )
+        # fetch user from database
+        asession = await get_async_session_direct()
+        try:
+            user_db = await get_user_by_username(username, asession)
+            await asession.aclose()
+            return user_db
+        except Exception as err:
```
sidravi1, line 100: same here. Let's not catch base exceptions

core_backend/app/users/models.py
```diff
@@ -90,12 +111,16 @@ async def get_user_by_username(
         return user
     except NoResultFound as err:
         raise ValueError(f"User with username {username} does not exist.") from err
+    except MultipleResultsFound as err:
```
sidravi1, line 114: Why should this ever happen? Can we make username unique in the Db?

core_backend/app/users/models.py
```diff
@@ -109,3 +134,7 @@ async def get_user_by_token(
         return user
     except NoResultFound as err:
         raise ValueError("User with given token does not exist.") from err
+    except MultipleResultsFound as err:
+        raise ValueError(
+            "Multiple users with given token found in local database."
```
sidravi1, line 139: Same here. Let's make the token unique in Db. Should never happen

core_backend/app/utils.py
```diff
@@ -15,6 +16,15 @@
 )
 
 
+def generate_key() -> str:
+    """
+    Generate a new 32-character API key
+    """
+
+    # 32 bytes gives 43 characters here
+    return secrets.token_urlsafe(32)
```
sidravi1, line 25: Should avoid magic numbers in code.

Don't need to put it in config but let's move it to the total as a global constant.
N_BYTES_SECRET_KEY or something similar.

core_backend/app/utils.py
```diff
@@ -15,6 +16,15 @@
 )
 
 
+def generate_key() -> str:
+    """
+    Generate a new 32-character API key
```
sidravi1, line 21: Is it 32 chars or 43 chars (as below)

core_backend/tests/api/conftest.py
```diff
@@ -42,6 +41,9 @@
 
 
 TEST_USER_ID = "test_user_id"
+TEST_USERNAME = "test_username"
```
sidravi1, line 44: Any advantage in turning these into a fixture? Maybe when you write tests?

core_backend/tests/api/test_manage_content.py
```diff
@@ -96,7 +96,7 @@ async def test_edit_and_retrieve_content(
         content_title: str,
         content_text: str,
         fullaccess_token: str,
-        # readonly_token: str,
+        # fullaccess_token: str,
```
sidravi1, line 99: delete commented line?

core_backend/tests/api/test_manage_content.py
```diff
@@ -116,7 +116,7 @@ async def test_edit_and_retrieve_content(
             f"/content/{existing_content_id}",
             headers={
                 "Authorization": f"Bearer {fullaccess_token}"
-            },  # used to be readonly_token
+            },  # used to be fullaccess_token
```
sidravi1, line 119: delete commented line?

core_backend/tests/api/test_manage_content.py
```diff
@@ -163,113 +163,6 @@ async def test_delete_content(
         assert response.status_code == 200
 
 
-class TestAuthManageContent:
```
sidravi1, line 166: I worry that we are removing readonly from tests but not fully from functionality. What are you thoughts?

core_backend/tests/api/test_manage_ud_rules.py
```diff
@@ -146,114 +145,6 @@ async def test_delete_UDrules(
         assert response.status_code == 200
 
 
-class TestAuthManageContent:
```
sidravi1, line 149: Were you able to check if all of this functionality is being tested elsewhere?

core_backend/tests/api/test_question_answer.py
```diff
@@ -22,12 +21,13 @@
...
     ResultState,
 )
+from core_backend.tests.api.conftest import TEST_USER_RETRIEVAL_KEY
 
 
 class TestEmbeddingsSearch:
     @pytest.mark.parametrize(
         "token, expected_status_code",
-        [(f"{USER1_RETRIEVAL_KEY}_incorrect", 401), (USER1_RETRIEVAL_KEY, 200)],
+        [(f"{TEST_USER_RETRIEVAL_KEY}_incorrect", 401), (TEST_USER_RETRIEVAL_KEY, 200)],
```
sidravi1, line 30: Probably best practice to a fixture if this is being imported within a test. In case a test changes the value etc. but you can do this in a the tests PR

sidravi1 (approved):
Thanks @amiraliemami! Approving since it is an intermediary PR but there are a few things to fix before merge into main.

amiraliemami:
Made changes based on review and responded to comments, thanks!

core_backend/app/database.py
amiraliemami: Thanks for the clarification!

core_backend/app/auth/dependencies.py
amiraliemami: I used get_sqlalchemy_async_engine() directly in this code with a `with` phrase - using generator-based functions was a headache here. Does this work? If you don't agree with the latest iteration then maybe we can cowork on it

amiraliemami:
New comments

amiraliemami (commented):
Latest comments

core_backend/tests/api/test_manage_ud_rules.py
sidravi1: Would you mind confirming?

sidravi1: Ok! I'm happy if you have confirmed this.

core_backend/tests/api/test_question_answer.py
sidravi1: You may want to test this with different RETRIEVAL_KEYS in the future. Like use USER1's to access USER2's content etc... Having a fixture will let you do that easily.

And it also gets around the problem of a test changing a global that is used by other tests - decouples tests.

core_backend/app/auth/dependencies.py
```diff
@@ -99,36 +84,31 @@ def get_current_user(
...
-            access_level=cast(AccessLevel, USERS[username]["access_level"]),
-        )
-
+        # fetch user from database
+        async with AsyncSession(
+            get_sqlalchemy_async_engine(), expire_on_commit=False
+        ) as asession:
+            try:
+                user_db = await get_user_by_username(username, asession)
+                await asession.aclose()
```
sidravi1, line 93: Don't need close anymore. The context manager sorts this out

core_backend/app/auth/dependencies.py
```diff
@@ -99,36 +84,31 @@ def get_current_user(
...
+                return user_db
+            except UserNotFoundError as err:
+                await asession.aclose()
```
sidravi1, line 96: Same here

core_backend/app/auth/dependencies.py
```diff
@@ -8,83 +8,68 @@
...
-
-def get_current_user(
-    token: Annotated[str, Depends(oauth2_scheme)]
-) -> AuthenticatedUser:
+    async with AsyncSession(
+        get_sqlalchemy_async_engine(), expire_on_commit=False
+    ) as asession:
+        try:
+            user_db = await get_user_by_username(username, asession)
+            await asession.aclose()
```
sidravi1, line 62: Not needed anymore!

core_backend/app/auth/dependencies.py
amiraliemami: Woops, will clean these up

PR #198, opened 2024-05-06
HOTFIX: fixed bug when click away from new items

sidravi1:
Reviewer: @amiraliemami 
Estimate: 5mins

Fixed this bug:

https://github.com/IDinsight/aaq-core/assets/14125957/119fb2dc-6cb2-4106-99c6-091c96ac1930

### DIFF
--- a/admin_app/src/app/urgency-rules/page.tsx
+++ b/admin_app/src/app/urgency-rules/page.tsx
@@ -120,11 +120,7 @@ const UrgencyRulesPage = () => {
   };
 
   const onBlur = (index: number) => {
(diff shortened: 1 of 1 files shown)


amiraliemami (approved):
Thanks for this quick fix! I should've suggested a behaviour when I raised the issue but the ideal thing I think would be for the text box to just stay there and wait for the user to click in it again... This fix makes it disappear (and so the work is lost because of a misclick).

We should merge this for now since it's definitely less buggy than before but could you make a ticket for improving its behaviour? Thanks!

PR #195, opened 2024-05-06
Multi-user support + Google login!

amiraliemami:
Reviewer: @suzinyou 
Estimate: ?

---

## Ticket

Fixes: [519](https://idinsight.atlassian.net/browse/AAQ-519?atlOrigin=eyJpIjoiMWIyZGE2ODI1NTNkNDI2M2IwMWEyMjIxNzJkMDQxMDEiLCJwIjoiaiJ9)

### Goal

- Make app usable by multiple users at once with different logins
- Add Google login
- Add the user the ability to generate and rotate their API key

### Changes

1. New users table which stores hashed API keys
2. New user_id column in other relevant tables
3. Content/UD rules are now added with the user_id of the logged-in user
4. API keys route retrieval/urgent detection to only use the contents/rules of the user who that API key belongs to
5. Added functionality for a user to generate/rotate their API key
6. Major db table name cleanup
7. alembic migrations cleanup
8. Read-only access removed from backend (all users are fullaccess)
9. Updated pytests to catch for inter-user data leakage
10. Added Google login ability (@sidravi1)
11. Updated AWS terraform and bootstrap.sh file with new env variables (to-be-merged)

It's been a journey.

### Future Tasks (optional)

- Create ability to limit number of contents each user is allowed to add and rate limit API use
- Add mention of this feature in the docs and blog post

## How has this been tested?

- run dev setup and play around
- run local deployment and play around
- run pytests with `make tests`

## To-do before merge (optional)

- Merge #217 into this branch

## Checklist

Fill with `x` for completed. Delete any lines that are not relevant

- [x] My code follows the style guidelines of this project
- [x] I have reviewed my own code to ensure good quality
- [x]  I have tested the functionality of my code to ensure it works as intended
- [x] I have resolved merge conflicts
- [x] I have updated the automated tests (if applicable)
- [x] I have updated the requirements (if applicable)
- [ ] I have updated affected documentation (if applicable)

PR #201, opened 2024-05-09
[AAQ-482] Add custom embeddings as optional components

lickem22:
Reviewer: @suzinyou 
Estimate: 1h

---

## Ticket

Fixes: [AAQ-482](https://idinsight.atlassian.net/browse/AAQ-482)

### Goal
The goal of this PR is to add local embeddings as an optional component. 

### Changes
- Added an optional app to load and deploy custom HuggingFace embeddings using Docker
- Added a make command to build an image for the embeddings app and start a container
- Added a docker compose component to 

### How to test:
- Add the following environment variables:
```
export EMBEDDINGS_API_ENDPOINT=http://host.docker.internal:8080
export EMBEDDINGS_API_KEY=add-token
export LITELLM_MODEL_EMBEDDING=openai/local-embeddings
export HUGGINGFACE_MODEL = thenlper/gte-large  #can be changed
```
- Start embedding container by doing:
`make setup-embeddings`
- Run the backend app. 

Now you can try to add content and ask questions using the custom embeddings
## How has this been tested?
Tested the embeddings app in the backend

If we want to merge this to main, we need to make sure we use the new openai embeddings allowing to take a subset of  the vector with a length of 1024
## Checklist

[AAQ-482]: https://idinsight.atlassian.net/browse/AAQ-482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


optional_components/embeddings/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+ENV PYTHONPATH=/app
+
+COPY ./requirements.txt /app/requirements.txt
+
+RUN pip install --no-cache-dir --upgrade -r /app/requirements.txt
+
+COPY ./app /app
+
+
+RUN ls -la /app
```
suzinyou, line 14: Is this just sanity check? Shall we remove?

optional_components/embeddings/app/main.py
```diff
@@ -0,0 +1,85 @@
...
+        "index": 0,
+        "embedding": embeddings.tolist()[0],
+    }
+    return ResponseModel(data=[response_dict])
+
+
+@app.post("/chat/completions", response_model=ResponseModel)
+async def get_completions(token: str = Depends(verify_token)):
+    """Get chat completions. Required for LiteLLM endpoints"""
+    return JSONResponse(status_code=200, content={"status": "ok"})
```
suzinyou, line 85: This we don't need right?

optional_components/embeddings/requirements.txt
```diff
@@ -0,0 +1,5 @@
+fastapi==0.109.1
+numpy==1.26.4
+torch==2.3.0
+transformers==4.40.1
+uvicorn==0.29.0
```
suzinyou, line 5:
```suggestion
uvicorn==0.29.0

```

optional_components/embeddings/app/config.py
```diff
@@ -0,0 +1,4 @@
+import os
+
+HUGGINGFACE_MODEL = os.environ.get("HUGGINGFACE_MODEL", "thenlper/gte-large")
+STATIC_TOKEN = os.environ.get("EMBEDDINGS_API_KEY", "add-token")
```
suzinyou, line 4: Can we call the config itself API_KEY just for clarity?

optional_components/embeddings/app/config.py
```diff
@@ -0,0 +1,4 @@
+import os
+
+HUGGINGFACE_MODEL = os.environ.get("HUGGINGFACE_MODEL", "thenlper/gte-large")
```
suzinyou, line 3: Oh so if we pass when running
```
env: 
  - HUGGINGFACE_MODEL: "someother/model"
```
we can switch up which model we use!

core_backend/migrations/versions/72d3851d44a6_create_ud_tables.py
suzinyou, line 1: Oo you might want to rebase after #195 is merged.. that changed the migration scripts

suzinyou: Oh but it's not merged so whoever merges first wins :D

suzinyou (commented):
Nice!!

optional_components/embeddings/app/config.py
lickem22: Yes. As long as it is an embedding model.

optional_components/embeddings/app/config.py
lickem22: You mean instead of `EMBEDDINGS_API_KEY`. I added it to be more specific since in the future we might have many api keys.

optional_components/embeddings/app/main.py
lickem22: It was part of the documentation. But I can check whether we can do it without or not.

lickem22: Turns out I don't need it. Will remove

suzinyou:
With testing, I also need to start the db right?

Can you add the instructions on the docs as well? Exciting new component!

core_backend/app/config.py
```diff
@@ -22,7 +22,9 @@
 # "model_name" in the proxy config.yaml.
 # "openai/..." is needed since the proxy presents a unified OpenAI-style API
 # for all of its endpoints.
-LITELLM_MODEL_EMBEDDING = os.environ.get("LITELLM_MODEL_EMBEDDING", "openai/embeddings")
+LITELLM_MODEL_EMBEDDING = os.environ.get(
+    "LITELLM_MODEL_EMBEDDING", "openai/local-embeddings"
```
suzinyou, line 26: I don't think we want to use local-embeddings by default. Requires setting up an optional component! Let's change this back to embeddings? Also see my comment in the litellm config

deployment/docker-compose/litellm_proxy_config.yaml
```diff
@@ -3,6 +3,11 @@ model_list:
     litellm_params:
       model: text-embedding-3-small
       api_key: "os.environ/OPENAI_API_KEY"
+  - model_name: local-embeddings
+    litellm_params:
+      model: openai/custom-embedding
+      api_base: "os.environ/EMBEDDINGS_ENDPOINT"
+      api_key: "os.environ/EMBEDDINGS_API_KEY"
```
suzinyou, line 10: I think the idea with this config file was to switch the models we use for each task, so /embeddings /detect-language / safety, etc.

As in, this is the file where we define the models. The env vars are there just for extra flexibility.

So we should just replace the current `model_name: embeddings` to have this config, not create a new model.

(We can switch the aws deployment config to this as well once we 1. host it 2. test it, and are happy with the performance)

optional_components/embeddings/app/main.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@app.on_event("startup")
+def load_model():
+    """Load the model and tokenizer on startup"""
+    global tokenizer
+    global model
+    tokenizer = AutoTokenizer.from_pretrained(HUGGINGFACE_MODEL)
+    config = AutoConfig.from_pretrained(HUGGINGFACE_MODEL)
+    model = AutoModel.from_pretrained(HUGGINGFACE_MODEL, config=config)
```
suzinyou, line 22: Is there a way to only run this if we haven't mounted the files?
Generally do we want to download always from huggingface hub or allow users to mount a volume containing their own embeddings?

We can create a ticket and discuss this in the future!

suzinyou (changes requested):
Thank you @lickem22 !!

How do I test the docker-compose? Using our docs our standard way to spin it up is 

```
docker compose -f docker-compose.yml -f docker-compose.dev.yml -p aaq-stack watch
```

But it doesn't spin up embeddings container..

core_backend/app/config.py
lickem22: I added it for easy testing since the embeddings default size is 1536. I will put it back to the default value when we will fix the embeddings size on main.

deployment/docker-compose/litellm_proxy_config.yaml
lickem22: Isn't the point of the proxy to have many models configured so we only have to change the name of the model we want? Because I think your approach actually ends up being less flexible.

optional_components/embeddings/app/main.py
lickem22: Yes there are several approaches to this, I used this approach because it is easy to build on docker. Next steps would be:
- Update the Dockerfile to download a model from anywhere 
- Copy to the environment
- Use the local model when deployed.

core_backend/app/config.py
suzinyou: Not sure I understand -- can you elaborate on this?

Also, I don't think we want to swap embeddings with gte-large just yet. We should evaluate some other options and switch?

deployment/docker-compose/litellm_proxy_config.yaml
suzinyou: Hmm maybe you're right. 

I thought that the whole point of the config was to define 1 `model_name` per prediction task (detect-language, safety, paraphrase, ...). And, each AAQ instance (for use case A, B, C, ..) would have their own proxy server with their own configs. 

But maybe the idea was to (have the option to) serve only 1 proxy server with many different model options for many AAQ instances, and each AAQ instance chooses which model they use by referring to the correct model_name as defined in the config?

cc: @amiraliemami

suzinyou: Can we call it huggingface_embeddings? It does not have to be local, it can be on cloud, but just that we can serve opensource embeddings, specifically hf it seems

core_backend/app/config.py
suzinyou: Let's discuss this on slack first

deployment/docker-compose/litellm_proxy_config.yaml
amiraliemami: The idea is to have our app code be stable (i.e. always call "openai/embeddings") and we just change the proxy's forwarding to a different model if we want. So yeah, should leave our AAQ code same as before and just change the routing API url and key to be smt else.

core_backend/app/config.py
lickem22: I put the embeddings size back to 1536

deployment/docker-compose/litellm_proxy_config.yaml
lickem22: I will do as we discussed on slack

optional_components/embeddings/app/main.py
```diff
@@ -0,0 +1,78 @@
...
+from pydantic import BaseModel
+from torch import Tensor
+from transformers import AutoConfig, AutoModel, AutoTokenizer
+
+app = FastAPI()
+
+security = HTTPBearer()
+
+
+@app.on_event("startup")
```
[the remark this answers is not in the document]
lickem22: I didn't see that warning but it's fixed!

lickem22:
> Thank you @lickem22 !!
> 
> How do I test the docker-compose? Using our docs our standard way to spin it up is
> 
> ```
> docker compose -f docker-compose.yml -f docker-compose.dev.yml -p aaq-stack watch
> ```
> 
> But it doesn't spin up embeddings container..

You can just do that:
```
docker-compose -f docker-compose.yml -f docker-compose.dev.yml -p aaq-stack --profile embeddings watch
```

lickem22:
> Can you add the testing/setup instructions on the docs as well? Exciting new component!
> 
> Also, the testing instructions are missing the following:
> 
> 1. Set up DB
> 2. Set up Litellm proxy server
> 3. The last export statement should be `export HUGGINGFACE_MODEL=thenlper/gte-large` (there shuoldn't be spaces around `=`!)

Hi Suzin, I am not sure what you mean by testing/setup instructions on the docs. Can you clarify? I will be  happy to update it in my PR for the blog (and probably docs) update.

suzinyou:
> > Can you add the testing/setup instructions on the docs as well? Exciting new component!
> > Also, the testing instructions are missing the following:
> > 
> > 1. Set up DB
> > 2. Set up Litellm proxy server
> > 3. The last export statement should be `export HUGGINGFACE_MODEL=thenlper/gte-large` (there shuoldn't be spaces around `=`!)
> 
> Hi Suzin, I am not sure what you mean by testing/setup instructions on the docs. Can you clarify? I will be happy to update it in my PR for the blog (and probably docs) update.

In our documentation, we should have a guide on how to use this optional component like we have for align score!

core_backend/app/contents/config.py
```diff
@@ -1,7 +1,7 @@
 import os
 
 # PGVector variables
-PGVECTOR_VECTOR_SIZE = os.environ.get("PGVECTOR_VECTOR_SIZE", "1536")
+PGVECTOR_VECTOR_SIZE = os.environ.get("PGVECTOR_VECTOR_SIZE", "1024")
```
suzinyou, line 4: Is this for gte-large? 
We don't define this env var in our deployment, which uses openai embeddings. So if the deployment just uses default value it may break?

suzinyou: Can we keep it as original value and add to the docs that we should update this env var to match whatever embeddings we use?

Also realized we should either 
1. never allow switching embedding models
2. or allow switching, but also have some way to trigger reindexing of all content to that embedding.. 😅 future ticket?

lickem22: For now I will update back to 1536

suzinyou: That's good to know. Generally we need an easier way to update these environment variables for deployments & updating vector db index..

suzinyou:
> > > Can you add the testing/setup instructions on the docs as well? Exciting new component!
> > > Also, the testing instructions are missing the following:
> > > 
> > > 1. Set up DB
> > > 2. Set up Litellm proxy server
> > > 3. The last export statement should be `export HUGGINGFACE_MODEL=thenlper/gte-large` (there shuoldn't be spaces around `=`!)
> > 
> > 
> > Hi Suzin, I am not sure what you mean by testing/setup instructions on the docs. Can you clarify? I will be happy to update it in my PR for the blog (and probably docs) update.
> 
> In our documentation, we should have a guide on how to use this optional component like we have for align score!

Hey @lickem22  do you think you can manage to add the documentation? You have most of the info already in this PR & my comments so I'm hoping we can add that in!

suzinyou (approved):
Thanks so much!!!

PR #205, opened 2024-05-13
Google login

sidravi1:
Reviewer: @amiraliemami 
Estimate: 1hr

---

## Ticket

Fixes: [AAQ-514](https://idinsight.atlassian.net/browse/AAQ-514)

### Goal

Allows login using Google Auth

### Changes

- Added `Sign in with Google` button
- New backend API to handle login and token generation
- Minor changes to schema

### Future Tasks (optional)

- Set `access_type` or user type to "trial" when logging through google
- Restrict # of calls based on account type
- Restrict # of content cards based on account type
- Publish account

## How has this been tested?

- Login using your work idinsight account.
- CRUD content
- CRUD UD
- Try playground
- Login using your personal google account
- CRUD content
- CRUD UD
- Try playground

**NOTE: Only our email addresses are allowed rn. To allow anyone to use their google account, we need to publish this application**

## Checklist

[AAQ-514]: https://idinsight.atlassian.net/browse/AAQ-514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core_backend/app/auth/config.py
```diff
@@ -3,3 +3,6 @@
 ACCESS_TOKEN_EXPIRE_MINUTES = os.environ.get("ACCESS_TOKEN_EXPIRE_MINUTES", 60 * 24 * 7)
 JWT_ALGORITHM = os.environ.get("JWT_ALGORITHM", "HS256")
 JWT_SECRET = os.environ.get("JWT_SECRET", "jwt-secret")
+NEXT_PUBLIC_GOOGLE_LOGIN_CLIENT_ID = os.environ.get(
```
sidravi1, line 6: Make sure to set your client! See slack message for client

sidravi1: https://idinsight.slack.com/archives/C05RSDLN1GA/p1715630159446209

amiraliemami:
Thanks! Just tried this quickly before I review properly. Two initial comments:

1. The google client ID seems to have issues, when it redirects I just get a blank page so I can't test the google functionality right now, It's actually a 400 error from Google with the message `Parameter client_id is not set correctly.` I set it as per your slack message.
2. What's the rational behind making the `user_id` an integer? We want to use uuids for all IDs so we're not susceptible to enumeration attacks where other user_ids can be guessed if you know one. Ideally we should also change all other IDs to not be integers. If it really has to be an integer,  then we also need to update the values in `core_backend/add_users_to_db.py` too - currently that's broken because it tries to insert a uuid.hex string. My priors are p strong about the fact that we should be using random strings as IDs though.

sidravi1:
> Thanks! Just tried this quickly before I review properly. Two initial comments:
> 
> 1. The google client ID seems to have issues, when it redirects I just get a blank page so I can't test the google functionality right now, It's actually a 400 error from Google with the message `Parameter client_id is not set correctly.` I set it as per your slack message.

Hmm... is this a local setup? Did you set it for both the terminals running admin app and backend?
If so, let's debug together tomorrow.

> 2. What's the rational behind making the `user_id` an integer? We want to use uuids for all IDs so we're not susceptible to enumeration attacks where other user_ids can be guessed if you know one. Ideally we should also change all other IDs to not be integers. If it really has to be an integer,  then we also need to update the values in `core_backend/add_users_to_db.py` too - currently that's broken because it tries to insert a uuid.hex string. My priors are p strong about the fact that we should be using random strings as IDs though.

I changed it so that it would auto-increment and we don't have to define it for each user.
The user never actually sends the `user_id` right? it's baked into the token that only the backend can decrypt. So no one really knows their userid and even if you guessed it, there is nothing you can do about it since you can decrypt the token.

But I may be missing something and happy to be convinced otherwise. Can you explain why your prior are so strong? Here's what I found :
1. https://www.reddit.com/r/PostgreSQL/comments/1acojhp/incrementing_id_vs_random_string_performance_best/
2. https://stackoverflow.com/questions/2551776/database-design-primay-key-id-vs-string
3. https://chat.openai.com/share/6a7f2050-4a0e-4b9b-b364-413a6fdd84cb

amiraliemami:
> Hmm... is this a local setup? Did you set it for both the terminals running admin app and backend? If so, let's debug together tomorrow.

Oh I only set it for backend in dev setup. I just set it for frontend as well and now the page loads but I get `Invalid Google credentials` when I try one of my accounts.

I had also tried with docker compose which didn't load the Google auth page, and all that should need is the client ID variable in the `.env` right?

amiraliemami:
> I changed it so that it would auto-increment and we don't have to define it for each user. The user never actually sends the `user_id` right? it's baked into the token that only the backend can decrypt. So no one really knows their userid and even if you guessed it, there is nothing you can do about it since you can decrypt the token.
> 
> But I may be missing something and happy to be convinced otherwise. Can you explain why your prior are so strong? Here's what I found :
> 
> 1. https://www.reddit.com/r/PostgreSQL/comments/1acojhp/incrementing_id_vs_random_string_performance_best/
> 2. https://stackoverflow.com/questions/2551776/database-design-primay-key-id-vs-string
> 3. https://chat.openai.com/share/6a7f2050-4a0e-4b9b-b364-413a6fdd84cb

I see, you're right it's always encrypted so not exposed. Given the performance gains/readability then I guess it makes sense to have it be an integer. Could you update the ID values in `core_backend/add_users_to_db.py` to be ints too then?

What do you think about changing the content and query IDs to uuids though? They're exposed as part of our API calls openly.

DjSan4os (approved):

sidravi1:
> Oh I only set it for backend in dev setup. I just set it for frontend as well and now the page loads but I get `Invalid Google credentials` when I try one of my accounts.

Hmm... let me make sure I typed in your email address correctly.

> I had also tried with docker compose which didn't load the Google auth page, and all that should need is the client ID variable in the `.env` right?

Yep! That's all you need. I'll look into it and get back to you.

amiraliemami (approved):
Thanks a lot for this, works great and is very clean! 

One thing to note: there's not way to authenticate as a google user in the backend, meaning that we can't test our endpoints in the swaggerui as a google user. Front-end can call backend fine because it goes through the full google auth cycle but backend doesn't have that. 

This means that right now it's not possible to test retrieval for google users because there's no way to make them a retrieval key.

amiraliemami:
Oh two more thoughts:

1. Could you update the docs to say that the client ID env var must be set and where
2. Could you make whether the Google login button appears dependent on whether that variable exists? Should be easy I imagine. Means we won't have non-functioning components if we decide not to enable it for a client, for example.

DjSan4os (approved):

DjSan4os (commented):
Djsan4os

sidravi1:
> Oh two more thoughts:
> 
> 1. Could you update the docs to say that the client ID env var must be set and where
> 2. Could you make whether the Google login button appears dependent on whether that variable exists? Should be easy I imagine. Means we won't have non-functioning components if we decide not to enable it for a client, for example.

On #1 - we don't do that for any of the env vars. All we say is to take a copy of template.env and set the vars. I think that's sufficient for now.
On #2 - Yep. Makes sense! I'll make the change

sidravi1:
> > Oh two more thoughts:
> > 
> > 1. Could you update the docs to say that the client ID env var must be set and where
> > 2. Could you make whether the Google login button appears dependent on whether that variable exists? Should be easy I imagine. Means we won't have non-functioning components if we decide not to enable it for a client, for example.
> 
> On #1 - we don't do that for any of the env vars. All we say is to take a copy of template.env and set the vars. I think that's sufficient for now. On #2 - Yep. Makes sense! I'll make the change

#2 is a little tricky given how (when) the Google Button Div gets rendered. I'm going to merge in for now and fix it later.

PR #206, opened 2024-05-14
Add user pytests

amiraliemami:
Reviewer: @suzinyou 
Estimate: 30mins

---

## Ticket

Fixes: [518](https://idinsight.atlassian.net/browse/AAQ-518?atlOrigin=eyJpIjoiMzQ1ZWM1YjU0ODJmNDFiYThmOGIyMWRlNjBkMjBiMTMiLCJwIjoiaiJ9)

### Goal

Add pytests for 
- the api key rotation endpoint
- user table functions
- content CRUD user overlaps
- UD CRUD
- user2 token denial for retrieval and UD detect

## How has this been tested?

- Activate the aaq environment and run `make tests` in core_backend
- Added where to find login credentials to dev docs

## Checklist

### DIFF
--- a/core_backend/tests/api/test_key_management.py
+++ b/core_backend/tests/api/test_key_management.py
@@ -1,0 +1,17 @@
+from fastapi.testclient import TestClient
+
+from core_backend.tests.api.conftest import TEST_USER_RETRIEVAL_KEY
+
+
+class TestKeyManagement:
+    async def test_get_new_retrieval_key(
+        self, client: TestClient, fullaccess_token: str
+    ) -> None:
+        response = client.put(
+            "/key",
+            headers={"Authorization": f"Bearer {fullaccess_token}"},
+        )
+
+        assert response.status_code == 200
+        json_response = response.json()
+        assert json_response["new_retrieval_key"] != TEST_USER_RETRIEVAL_KEY
--- a/core_backend/tests/api/test_key_management.py
+++ b/core_backend/tests/api/test_key_management.py
@@ -1,17 +1,17 @@
 from fastapi.testclient import TestClient
 
-from core_backend.tests.api.conftest import TEST_USER_RETRIEVAL_KEY
+from core_backend.tests.api.conftest import TEST_USER_RETRIEVAL_KEY_2
 
 
 class TestKeyManagement:
     async def test_get_new_retrieval_key(
-        self, client: TestClient, fullaccess_token: str
+        self, client: TestClient, fullaccess_token_user2: str
     ) -> None:
         response = client.put(
             "/key",
-            headers={"Authorization": f"Bearer {fullaccess_token}"},
+            headers={"Authorization": f"Bearer {fullaccess_token_user2}"},
         )
 
         assert response.status_code == 200
         json_response = response.json()
-        assert json_response["new_retrieval_key"] != TEST_USER_RETRIEVAL_KEY
+        assert json_response["new_retrieval_key"] != TEST_USER_RETRIEVAL_KEY_2
--- a/core_backend/tests/api/test_manage_content.py
+++ b/core_backend/tests/api/test_manage_content.py
@@ -159,6 +159,8 @@ class TestManageContent:
         )
         assert response.status_code == 200
 
+
+class TestMultUserManageContent:
     @pytest.mark.parametrize(
         "content_title, content_text, content_metadata",
         [("title 3", "test content 3", {})],
--- a/core_backend/tests/api/test_manage_content.py
+++ b/core_backend/tests/api/test_manage_content.py
@@ -161,68 +161,47 @@ class TestManageContent:
 
 
 class TestMultUserManageContent:
-    @pytest.mark.parametrize(
-        "content_title, content_text, content_metadata",
-        [("title 3", "test content 3", {})],
-    )
-    async def test_content_access_overlap(
+    async def test_user2_get_user1_content(
         self,
         client: TestClient,
-        content_title: str,
-        content_text: str,
-        fullaccess_token: str,
+        existing_content_id: str,
         fullaccess_token_user2: str,
-        content_metadata: Dict[Any, Any],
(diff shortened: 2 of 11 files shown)


amiraliemami:
Hey @sidravi1 @suzinyou  I’ve done my best on adding pytests for the multi-user feature. 

- All my tests work now but the way async is being handled is making some UD tests fail and I can’t figure it out. Could you guys have a look and let me know if you can find out why it’s failing? 
- Also, I don’t have much experience with fixtures and pytests so it’s likely there’s a better way to handle the manual user additions that are done right now.

core_backend/tests/api/test_key_management.py
```diff
@@ -0,0 +1,17 @@
+from fastapi.testclient import TestClient
+
+from core_backend.tests.api.conftest import TEST_USER_RETRIEVAL_KEY_2
+
+
+class TestKeyManagement:
+    async def test_get_new_retrieval_key(
```
suzinyou, line 7: This is my first time reviewing multiuser feature.. so sorry if I missed some context here, but is this key just for /embeddings-search? Or also for llm generated responses & UD? Thought "retrieval" would be confusing if we can do all three operations with the same key!

core_backend/tests/api/test_manage_content.py
```diff
@@ -160,6 +160,71 @@ async def test_delete_content(
         assert response.status_code == 200
 
 
+class TestMultUserManageContent:
+    @pytest.mark.parametrize(
+        "content_title, content_text, content_metadata",
+        [("title 3", "test content 3", {})],
+    )
```
suzinyou, line 167: parametrize makes sense if we want to test multiple different combos of parameters, but
seems like we're just using one! Fine to keep it though I think

core_backend/tests/api/test_manage_content.py
```diff
@@ -160,6 +160,71 @@ async def test_delete_content(
...
+                "content_text": content_text,
+                "content_language": "ENGLISH",
+                "content_metadata": content_metadata,
+            },
+        )
+        assert response.status_code == 200
+        json_response = response.json()
+        assert json_response["content_metadata"] == content_metadata
+        assert "content_id" in json_response
+
```
suzinyou, line 192: Thanks for this!! May I suggest breaking down this tests however?
Each comment defines what the next block of code is testing... ideally each test would
be that block of code! Because if we fail "try and fail to fetch content as user2" we
have no idea whether the rest of the logic works or not.

I think you can use something like the `existing_content_id` fixture to achieve this.
Create a fixture like that using user 1, and reuse it for each unit test:
1. User 1 can delete it (*although I think this is tested by the TestManageContent class?)
2. User 2 can't get it
3. User 2 can't edit it
4. User 2 can't delete it

suzinyou, line 192: Thanks for this!! May I suggest breaking down this tests however?
Each comment defines what the next block of code is testing... ideally each test would
be that block of code! Because if we fail "try and fail to fetch content as user2" we
have no idea whether the rest of the logic works or not.

I think you can use something like the `existing_content_id` fixture to achieve this.
Create a fixture like that using user 1, and reuse it for each unit test:
1. User 1 can delete it (*although I think this is tested by the TestManageContent class?)
2. User 2 can't get it
3. User 2 can't edit it
4. User 2 can't delete it

core_backend/tests/api/test_manage_ud_rules.py
```diff
@@ -43,7 +43,7 @@ class TestManageUDRules:
         "urgency_rule_text, urgency_rule_metadata",
         [
             ("test rule 3", {}),
-            ("test rule 2", {"meta_key": "meta_value"}),
+            ("test rule 4", {"meta_key": "meta_value"}),
```
suzinyou, line 46: Shall I ignore this?

core_backend/tests/api/test_users.py
```diff
@@ -0,0 +1,64 @@
...
+class TestUsers:
+    @pytest.fixture(scope="function")
+    async def asession(self) -> AsyncGenerator[AsyncSession, None]:
+        async for session in get_async_session():
+            yield session
+
+        await session.close()
+
+    async def test_save_user_to_db(self, asession: AsyncSession) -> None:
+        user = UserCreate(username="test_username_3")
```
suzinyou, line 32: Are there rules around what usernames are allowed? Would that be worth testing in
another test?

suzinyou: It's a question not a suggestion!!

suzinyou (changes requested):
Thanks Amir!! Left some suggestions/questions. 

This is not part of this PR but I'm seeing this deprecation warning:

> The method "utcnow" in class "datetime" is deprecated
>  Use timezone-aware objects to represent datetimes in UTC; e.g. by calling .now(datetime.UTC)

core_backend/tests/api/test_users.py
amiraliemami: Nope no rules about usernames yet!

core_backend/tests/api/test_manage_ud_rules.py
amiraliemami: Yup

core_backend/tests/api/test_key_management.py
amiraliemami: Yes you're right, when I renamed it to that we didn't have UD yet. Will change it to api_key in a separate final PR before merging mulittenancy to main, thanks!

amiraliemami:
> This is not part of this PR but I'm seeing this deprecation warning:
> 
> > The method "utcnow" in class "datetime" is deprecated
> > Use timezone-aware objects to represent datetimes in UTC; e.g. by calling .now(datetime.UTC)

Yup, been ignoring it lol added a backlog ticket to fix this

suzinyou (approved):

PR #221, opened 2024-05-30
Add tags landing

lickem22:
Reviewer:@sidravi
Estimate: 1h

---

## Ticket

Fixes: [AAQ-549](https://idinsight.atlassian.net/browse/AAQ-549)

### Goal
Th Goal of this PR is to allow filtering of contents using content tags and allow to add new tags within the tags bar. 
### Changes
Changes on the backend: No changes
Changes on the front-end:
- Added the Tags Bar on the landing page
- Added filtering feature using the Tag bar
- Updated the Tags Bar on the edit page to add a new tag when the text in the search doesn't return anything. Just like this:
Screenshot 2024-05-30 at 17 45 47
Screenshot 2024-05-30 at 17 50 00

### Future Tasks (optional)
- Edit and delete tags
## How has this been tested?
This has been tested by:
- Adding tags to the DB using the backend (Tags need to be added first otherwise the Tags Bar wont be displayed
- Added new contents with tags
- Edited contents by adding/editing tags 
- Tested filtering on landing page
- Tested adding new tag with add "tag"button. 
I suggested testing the feature like this to make sure everything works as intended
## Checklist

- [ ] I have added blogpost in Latest Updates


admin_app/src/app/content/edit/page.tsx
```diff
@@ -224,6 +240,31 @@ const ContentBox = ({
...
         )}
+        filterOptions={(options, params) => {
+          const filtered = filter(options, params);
+
+          const { inputValue } = params;
+          const isExisting = options.some(
+            (option) => inputValue === option.tag_name
+          );
+          if (inputValue !== "" && !isExisting) {
+            filtered.push({ tag_id: 0, tag_name: `Add "${inputValue}"` });
```
amiraliemami, line 251: This is awesome! Can we somehow make sure the tag stays in original form and is not forced to be in caps?

core_backend/app/tags/models.py
```diff
@@ -18,7 +18,7 @@
 from .schemas import TagCreate
 
 content_tags_table = Table(
-    "content_tags",
+    "content-tag",
```
amiraliemami, line 21: Hey @lickem22, this breaks other relationships - just tried to deploy and getting errors when trying to save content

admin_app/src/app/content/edit/page.tsx
```diff
@@ -224,6 +240,31 @@ const ContentBox = ({
             placeholder="Add Tags"
           />
         )}
+        filterOptions={(options, params) => {
```
amiraliemami, line 243: I  see that the filtering seems to be an AND, let's do an OR instead as you said? To me showing all contents that have either tag selected in the filter box makes more intutive sense

admin_app/src/app/content/page.tsx
```diff
@@ -128,13 +179,22 @@ const CardsGrid = ({
...
             card.content_title
               .toLowerCase()
               .includes(searchTerm.toLowerCase()) ||
-            card.content_text.toLowerCase().includes(searchTerm.toLowerCase()),
-        );
+            card.content_text.toLowerCase().includes(searchTerm.toLowerCase());
+
+          const matchesAllTags = filterTags.every((fTag) =>
+            card.content_tags.includes(fTag.tag_id)
+          );
```
amiraliemami, line 191: I  see that the filtering seems to be an AND, let's do an OR instead as you said? To me showing all contents that have either tag selected in the filter box makes more intutive sense

admin_app/src/app/content/edit/page.tsx
```diff
@@ -224,6 +240,31 @@ const ContentBox = ({
...
+          return filtered;
+        }}
+        renderOption={(props, option) => {
+          if (option.tag_name) {
+            return (
+              <li {...props}>
+                <Button fullWidth onClick={() => handleNewTag(option.tag_name)}>
+                  {option.tag_name}
+                </Button>
+              </li>
```
amiraliemami, line 263: What do you think of this? Seems cleaner and gets rid of the weird button-inside-a-list effect

```suggestion
              <li {...props} onClick={() => handleNewTag(option.tag_name)}>
                  {option.tag_name}
              </li>
```

core_backend/app/tags/models.py
suzinyou: Instead of rewriting migration script we need a new migration that changes the table name from content_tag to content-tag!

admin_app/src/app/content/page.tsx
lickem22: I think we agreed on AND as far as I can remember. Let's discuss on Slack

core_backend/app/tags/models.py
lickem22: It's not supposed to be an issue because the model still reflects the revision. What erroe do you get? I doubt it is because of the migration script

suzinyou: When I add tags using UI I get this error on the backend:

> sys:1: SAWarning: Object of type <ContentDB> not in session, add operation along 'TagDB.contents' won't proceed (This warning originated from the Session 'autoflush' process, which was invoked automatically in response to a user-initiated operation.)

You are right that I didn't get any errors with the migration step when backend container started.. but why would we change the migration script? I'm not so experienced with DB migrations so just trying to understand. Isn't [core_backend/migrations/versions/2024_05_17_b5ad153a53dc_add_tags.py](https://github.com/IDinsight/aaq-core/pull/221/files#diff-bb14214a7eeaf41c3967d124ed02ddf8bf4d4284341c5043917851b67466e4ce) tied to the latest commit hash on main, and if we want to make any changes to the DB shouldn't we add another migration script?

lickem22: The goal of the scripts is to have versioning for the database schema just like git. So that when things break, you can revert easily.  Usually when you made a mistake in one of the scripts, you would delete the old migration file and recreate a new migration file. But I decided to update the migration script manually because the result would have been the same whether I generate it or update it manually. 
Also, the error you are getting, I am not able to replicate it

lickem22: I did some research and basically the reason we have the warning is because the ContentDB model was not added to the session yet. Recreating this warning is a bit tricky but I think I fixed it. Let me know if you still experience it.

sidravi1 (changes requested):
Hi @lickem22 - Thanks for this.

I'm having a few issues on my side:

1. I cannot use the mouse to select a tag from the drop down.
2. If i type it in and select it, it throws these warnings in the console:

image

3. If i create a new tag with the same name as an existing one, it throws an error

image

It should probably handle this cleanly. Maybe (a) don't allow it - remove the add option (b) say already exists

4. When i use the keyboard to select "Add" to create it, it gives something like this:

image

And doesn't actually create anything in the backend

5. What are your thoughts on creating uppercase and lowercase tags? Right now it allows you to create "hello" and "Hello". I think all tags should be lowercase only and it shouldn't allow the second one and handle it properly (see #3 above). But ok to be overridden if thats what the majority feel is better behaviour.

6. Since the "Save" button is in a different box, it is not obvious to me that you need to hit save. Can we have the state as dirty and throw a warning saying changes will be lost?

7. You probably want to add the props recommended here: https://mui.com/material-ui/react-autocomplete/#creatable

8. Would you also be able to close comment

lickem22:
> * What are your thoughts on creating uppercase and lowercase tags? Right now it allows you to create "hello" and "Hello". I think all tags should be lowercase only and it shouldn't allow the second one and handle it properly (see [Added Tests #3](https://github.com/IDinsight/aaq-core/pull/3) above). But ok to be overridden if thats what the majority feel is better behaviour.
> *

Thanks Sid.
1) Fixed
2) This is a current issue with MUI, I checked online but there is not a fixed for it yet. They are planning to remove it in MUI 6 since it is not a useful warning. (cf this [thread](https://github.com/mui/material-ui/issues/29727#issuecomment-2057764910))
3) Fixed. By checking that the value is not in both availableTags and contentTags
4) Also fixed
5) I think uppercase makes more sense. So now all tags are uppercase. 
6) Not sure I understand that. Are you suggesting we add a warning when leaving the page without saving the updated tags? If we do that for tags, shouldn't we do it for all unsaved updates on contents? Actually I think, if putting the tags outside the contentBox is the issue then we should put it inside. Maybe it is worth discussing with the team and I can update in another PR. 
7) done
8) Not sure I understand that but let me know if I still need to do it

admin_app/package-lock.json
lickem22, line 1: Updated material ui to fix the AutoComplete warning but didn't work. I didn't remove it because i think it does no harm to have the up to date version, but open to reverse the changes.

admin_app/src/app/content/edit/page.tsx
```diff
@@ -224,7 +249,45 @@ const ContentBox = ({
...
+            );
+          }
+          return (
+            <li key={option.tag_id} {...newProps}>
+              {option.tag_name}
+            </li>
+          );
+        }}
         sx={{ width: "500px" }}
+        isOptionEqualToValue={(option, value) =>
```
lickem22, line 288: It is supposed to fix the tags filter warning. But apparently there is a bug on MUI side. They plan to fix it in the next version

sidravi1:
@lickem22 

1. I think lowercase looks nicer but I'm ok to go with your choice.
2. When i select "Add" using my keyboard (down and enter), it doesn't actually create it but just says "Add "<tag>"". See below

image

leads to 

image

And an error when saving.

3. On #6 above. Yeah! We should do it for all content. Can you raise a ticket to track it? Should be an easy fix since we track it for enabling the "Save" button already.

sidravi1 (approved):
Thanks @lickem22 - Approving but please merge after you have fixed the issue below.

1. When i select "Add" using my keyboard (down and enter), it doesn't actually create it but just says "Add """. See below

image

leads to

image

And an error when saving.

2. #6 above. Yeah! We should do it for all content. Can you raise a ticket to track it? Should be an easy fix since we track it for enabling the "Save" button already.

core_backend/tests/api/test_manage_tags.py
```diff
@@ -70,8 +70,8 @@ async def test_create_and_delete_tag(
     @pytest.mark.parametrize(
         "tag_name",
         [
-            ("tag_first"),
-            ("tag_second"),
+            ("TAG_FIRST"),
```
sidravi1, line 73: We should test if that we cannot create "Tag1" if "tag1" or "TAG1" exist.

sidravi1: Maybe extend `test_add_tag_same_name_fails`

PR #228, opened 2024-06-06
Minor admin app improvements

sidravi1:
Reviewer: @lickem22 
Estimate: 20mins

---

## Ticket

N/A

### Goal

Alerts when there are unsaved changes

image

## How has this been tested?

From the front end

After #221 has been merge

## Checklist

- [ ] I have added blogpost in Latest Updates


admin_app/src/app/content/edit/page.tsx
```diff
@@ -68,7 +85,12 @@ const AddEditContentPage = () => {
   return (
     <FullAccessComponent>
       <Layout.FlexBox flexDirection={"column"} sx={{ p: sizes.doubleBaseGap }}>
-        <Header content_id={content_id} />
+        <Header
+          content_id={content_id}
+          onBack={() =>
+            isSaved ? router.push("/content") : setOpenDiscardChangesModal(true)
```
lickem22, line 91: I noticed something: When we make changes then discard them (whether it is adding a new tag or update the text), it still shows the modal. That can be confusing because then users would think they made changes even tho they didn't. I think we should find a way to make sure that when they content is at his original state after making changes, we can go leave the edit page.

lickem22 (approved):
Thanks Sid, Just one comment added. Nothing substential so it's approved.

admin_app/src/app/content/edit/page.tsx
sidravi1: Good point! We just track if anything changed. Instead we may want to do a diff between what was saved and new state. This is a good amount of logic to add and will park it for another day.

PR #231, opened 2024-06-06
Fix migrations content-tags issue

lickem22:
Reviewer: @suzinyou 
Estimate: 15mins

---

## Ticket

Fixes: [HOTFIX]

### Goal
Reverted migration update and added a new migration script instead. 
## Checklist

Fill with `x` for completed. 


core_backend/migrations/versions/2024_06_06_29b5ffa97758_update_content_tags.py
suzinyou, line 1: How about we just change the table name? This way we don't delete the data stored in original table?

https://www.programcreek.com/python/example/87984/alembic.op.rename_table

suzinyou (changes requested):
Let me know what you think! I'm not as well versed in migrations..

suzinyou:
Hi @lickem22 , this doesn't work. How did you test this? I realize that section is empty here.

This is how I tested it.
1. Check out at a commit on main before #221 was merged
2. Delete any existing DB volume and docker compose up. (in docker-compose.dev.yml, expose ports 5432:5432 to view the tables through, say, pgadmin)
3. Add some tags using Swagger UI
4. Now check out this branch
5. docker compose up. 
6. In backend_container, the migrations should run fine

But in step 6 I get this error
> sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedTable) relation "content-tags" does not exist
> [SQL: ALTER TABLE "content-tags" RENAME TO "content-tag"]

Indeed, on pgadmin I see that only content_tags table is there, not content-tags. So might be a matter of changing that in the script. 

Before re-requesting my review, could you please make sure the branch works as expected? And to ensure that this deploys properly, can you merge this branch to our testing environment and check that it deploys correctly? Let me know if you need help with that part!

suzinyou (changes requested):
Hi @lickem22 , this doesn't work. How did you test this? I realize that section is empty here.

This is how I tested it.

1. Check out at a commit on main before https://github.com/IDinsight/aaq-core/pull/221 was merged
2. Delete any existing DB volume and docker compose up. (in docker-compose.dev.yml, expose ports 5432:5432 to view the tables through, say, pgadmin)
3. Add some tags using Swagger UI
4. Now check out this branch
5. docker compose up.
6. In backend_container, the migrations should run fine

But in step 6 I get this error

> sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedTable) relation "content-tags" does not exist
> [SQL: ALTER TABLE "content-tags" RENAME TO "content-tag"]


Indeed, on pgadmin I see that only content_tags table is there, not content-tags. So might be a matter of changing that in the script.

Before re-requesting my review, could you please make sure the branch works as expected? And to ensure that this deploys properly, can you merge this branch to our testing environment and check that it deploys correctly? Let me know if you need help with that part!

lickem22:
> 6\. In backend_container, the migrations should run fine

Hi Suzin, have you tested it in dev mode before ? If it worked in dev mode but didn't in deployment mode it means it is not a code issue but a DB issue. You need to delete the DB volumes (ideally all volumes just in case) on docker then try again. You probably using old volumes that have content-tag instead of content-tags. It is a recurrent issue with docker-compose that I  flagged that a few weeks ago 
Try again and let me know

sidravi1:
> > 6. In backend_container, the migrations should run fine
> 
> Hi Suzin, have you tested it in dev mode before ? If it worked in dev mode but didn't in deployment mode it means it is not a code issue but a DB issue. You need to delete the DB volumes (ideally all volumes just in case) on docker then try again. You probably using old volumes that have content-tag instead of content-tags. It is a recurrent issue with docker-compose that I flagged that a few weeks ago Try again and let me know

The point of this PR is that you shouldn't have to clobber the db. You should be able to make incremental changes through migration scripts so you don't lose all the information. I think Suzin's process is right to recreate this issue. Also, may be worth getting on a quick huddle to resolve this.

lickem22:
> > > 6. In backend_container, the migrations should run fine
> > 
> > 
> > Hi Suzin, have you tested it in dev mode before ? If it worked in dev mode but didn't in deployment mode it means it is not a code issue but a DB issue. You need to delete the DB volumes (ideally all volumes just in case) on docker then try again. You probably using old volumes that have content-tag instead of content-tags. It is a recurrent issue with docker-compose that I flagged that a few weeks ago Try again and let me know
> 
> The point of this PR is that you shouldn't have to clobber the db. You should be able to make incremental changes through migration scripts so you don't lose all the information. I think Suzin's process is right to recreate this issue. Also, may be worth getting on a quick huddle to resolve this.

Hi Sid, I understand your point  but if the issue she is experiencing is on local development, then it is a docker issue not a migration issue. I don't know if you remember but I had the same issue when I tried deploying locally but was facing an issue where the state of the DB was on multilingual and although I was deploying main, the state didn't change and you were the one to tell me  I had to delete all volumes to make it work. 
On another hand, if it is one of the deployment then, it might be because that deployment was made with the update version of the migrations which uses `content-tag` instead of `content-tags`. If that's the case there is no other way but to revert the migration to before tags and upgrade.

suzinyou:
Thanks for the fix! will test soon.  We can go over why I tested it this way in a meeting. In reply to you points:

> You need to delete the DB volumes (ideally all volumes just in case) on docker then try again. 

In Step 2 I made sure the DB volume is created from scratch!

> You probably using old volumes that have content-tag instead of content-tags.

In step 6  "only content_tags table is there, not content-tags" and not "content-tag" either.

>  It is a recurrent issue with docker-compose that I flagged that a few weeks ago Try again and let me know

We should also talk about what issue you're facing with docker compose and fix it. Let's discuss that too!

suzinyou (approved):
Cool! The original test I described worked. And then I ran `alembic downgrade b5ad153a53dc` and then `alembic upgrade head` inside the backend container, and checked that the DB was updated properly.

PR #235, opened 2024-06-10
[Hotfix] set project name as module constant

suzinyou:
Reviewer: @amiraliemami 
Estimate: 15 min

---

## Ticket

Fixes: #234 

### Goal

Not exactly sure how it happens, but on deployed instance, the first generation in a trace was getting a random string prefix, hinting that `litellm.langFuseLogger` had not been initialized.

### Changes

We load the langfuse project name into a utils-module constant.

If litellm.langFuseLogger is not yet initialized, we initialize it as they do in https://github.com/BerriAI/litellm/blob/90ae7f040a904a6a183377e5ec8fdaf6d374a3e5/litellm/utils.py#L1862-L1865 . 

Gets rid of the random string prefix -- it would have put same generations into different traces.

## How has this been tested?

1. Locally ran `python main.py` in debug mode to see if langFuseLogger is None -- it is None initially
2. Made some /embeddings-search calls to make sure the fix always prefixes with project name when available.

## Checklist

Fill with `x` for completed. 

### DIFF
--- a/core_backend/app/utils.py
+++ b/core_backend/app/utils.py
@@ -22,6 +22,22 @@ from .config import (
 SECRET_KEY_N_BYTES = 32
 
 
+# To prefix trace_id with project name
+LANGFUSE_PROJECT_NAME = None
+
+if LANGFUSE == "True":
+    langFuseLogger = litellm.utils.langFuseLogger
+    if langFuseLogger is None:
+        langFuseLogger = litellm.integrations.langfuse.LangFuseLogger()
+        LANGFUSE_PROJECT_NAME = (
+            langFuseLogger.Langfuse.client.projects.get().data[0].name
+        )
+    elif isinstance(langFuseLogger, litellm.integrations.langfuse.LangFuseLogger):
+        LANGFUSE_PROJECT_NAME = (
+            langFuseLogger.Langfuse.client.projects.get().data[0].name
+        )
+
+
 def generate_key() -> str:
     """
     Generate API key (default 32 byte = 43 characters)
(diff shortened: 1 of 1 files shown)


core_backend/app/utils.py
```diff
@@ -22,6 +22,22 @@
...
+    langFuseLogger = litellm.utils.langFuseLogger
+    if langFuseLogger is None:
+        langFuseLogger = litellm.integrations.langfuse.LangFuseLogger()
+        LANGFUSE_PROJECT_NAME = (
+            langFuseLogger.Langfuse.client.projects.get().data[0].name
+        )
+    elif isinstance(langFuseLogger, litellm.integrations.langfuse.LangFuseLogger):
+        LANGFUSE_PROJECT_NAME = (
+            langFuseLogger.Langfuse.client.projects.get().data[0].name
+        )
```
suzinyou, line 38: Or just in case there's trouble getting the project name, I'm not sure if we should do something like

```suggestion
def get_random_string(size: int) -> str:
    """Generate a random string of fixed length."""
    import random
    import string

    return "".join(random.choices(string.ascii_letters + string.digits, k=size))


# To prefix trace_id with project name
LANGFUSE_PROJECT_NAME = None

if LANGFUSE == "True":
    langFuseLogger = litellm.utils.langFuseLogger
    try:
        if langFuseLogger is None:
            langFuseLogger = litellm.integrations.langfuse.LangFuseLogger()
        LANGFUSE_PROJECT_NAME = (
            langFuseLogger.Langfuse.client.projects.get().data[0].name
        )
    except Exception as e:
        LANGFUSE_PROJECT_NAME = get_random_string(8).lower()
```

amiraliemami (approved):
Lgtm :)

PR #238, opened 2024-06-11
Fixed Db connection and string

sidravi1:
Reviewer: @amiraliemami 
Estimate: 30mins

---

## Ticket

Fixes: JIRA_TICKET_LINK

### Goal

We had issues where db was reusing connection across different event loops.
We did a temporary fix by setting pooling to NullPool. This meant that we didn't keep a pool of connections and created a new connection for each request. 

But the downside is:
1. It is slower. There is an overhead for every connection
2. Once connections reach 100 (max connection right now on our postgres instance), the new ones are dropped
3. If we were using pooling, it would be added to the queue.

### Changes

We probably shouldn't change the core code to make tests work. This changes tests to make them work.
* Removed NullPool and made it 20 connections in pool
* Created engine again in conftest so that it's in the same event loop as the tests

## How has this been tested?

Tests pass

## Checklist

Fill with `x` for completed. 

### DIFF
--- a/core_backend/Makefile
+++ b/core_backend/Makefile
@@ -9,7 +9,7 @@ tests: setup-test-containers run-tests teardown-test-containers
 run-tests:
 	@set -a && source ./tests/api/test.env && set +a && \
 	python -m alembic upgrade head && \
-	python -m pytest -rPQ -m "not rails"
+	python -m pytest -rPQ -m "not rails" tests
 
 ## Helper targets
 setup-test-containers: setup-test-db
--- a/core_backend/tests/api/conftest.py
+++ b/core_backend/tests/api/conftest.py
@@ -2,7 +2,7 @@ import json
 import random
 from collections import namedtuple
 from datetime import datetime
-from typing import Any, Dict, Generator, List, Optional, Tuple
+from typing import Any, AsyncGenerator, Dict, Generator, List, Optional, Tuple
 
 import httpx
 import numpy as np
@@ -10,6 +10,7 @@ import pytest
 from fastapi.testclient import TestClient
 from pytest import Item
 from pytest_asyncio import is_async_test
+from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession, create_async_engine
 from sqlalchemy.orm import Session
 
 from core_backend.app import create_app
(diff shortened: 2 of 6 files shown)


core_backend/Makefile
```diff
@@ -9,7 +9,7 @@ tests: setup-test-containers run-tests teardown-test-containers
 run-tests:
 	@set -a && source ./tests/api/test.env && set +a && \
 	python -m alembic upgrade head && \
-	python -m pytest -rPQ -m "not rails"
+	python -m pytest -rPQ -m "not rails" tests
```
sidravi1, line 12: So pytest doesn't compile `validation` folder

core_backend/tests/api/conftest.py
```diff
@@ -308,8 +329,8 @@ async def mock_dashboard_stats(*arg: str, **kwargs: str) -> QuestionDashBoard:
     statistics.
     """
     return QuestionDashBoard(
-        six_months_question=[random.randint(0, 100) for _ in range(6)],
-        sis_months_upvote=[random.randint(0, 100) for _ in range(6)],
+        six_months_questions=[random.randint(0, 100) for _ in range(6)],
```
sidravi1, line 332: How in the world did tests pass with this error? We may need to check the tests for Dashboards
Also my IDE threw an error

suzinyou (approved):
Tests pass! Thank you!
But please check out my comments @sidravi1 @lickem22

core_backend/tests/api/conftest.py
sidravi1: Ah! looks like it was not mocking correctly and so was calling the actual endpoint. This was returning zeroes for both values and the test was passing.

@lickem22 - Can i suggest:
1. Move `mock_dashboard_stats` to `test_question_dashboard.py`. It doesnt need to be in conftest
2. It should take two arguments: questions, upvotes
3. Use functools.partial to create a function with these argument set to (0, 0), (1000, 1000), (1000, 0), (0, 1000) and anything else you think makes sense
4. mock `core_backend.app.question_dashboard.routers.get_dashboard_stats` with this partial
5. Check that returned values match the values set in #3

core_backend/tests/api/conftest.py
```diff
@@ -78,6 +79,26 @@ def db_session() -> Generator[Session, None, None]:
         next(session_gen, None)
 
 
+# We recreate engine and session to ensure it is in the same
+# event loop as the test. Without this we get "Future attached to different loop" error.
```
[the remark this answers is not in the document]
sidravi1: Test were already sharing the same event loop. What is was happening is that when we create a test client, it also creates an async db engine that is on a different event loop (fastapi event loop). So you can't use it in a pytest event loop.

core_backend/tests/api/conftest.py
lickem22: Hi Sid, I am not sure I understand 5. What values are set in the PR you linked ?

PR #242, opened 2024-06-12
Add tags script

lickem22:
Reviewer: @suzinyou 
Estimate: 30mins

---

## Ticket

Fixes: []()

### Goal
The goal of this PR is to add tags to the add content scripts. 
### Changes
- Update script with add tags logic

## How has this been tested?
By testing the script on dummy tags. 

## Checklist

Fill with `x` for completed. 

- [x] I have updated the scripts in `scripts/`


suzinyou (commented):
We need to fix a few things for this branch, but also let's follow up on slack about how to tackle this!

scripts/add_content_to_db.py
```diff
@@ -70,17 +70,37 @@
...
+        tags = set(tags)
+        tags_map = {}
+        for tag in tags:
+            payload = {
+                "tag_name": tag,
+            }
+            response = requests.post(tags_endpoint, json=payload, headers=headers)
+            if "tag_id" not in response.json():
+                print(f"Failed to create tag: {tag}")
+                continue
```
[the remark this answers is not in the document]
suzinyou: Ah I thought our POST /content endpoint checks for duplicate titles, but it seems it doesn't. 

The DB may not always be empty.. we might be using the upload script multiple times. And I feel a bit uncomfortable assuming DB will be empty bc script users/Tanmay probably doesn't have access to wipe the DB clean if something goes wrong. But I'll leave it up to you to discuss with Amir what's best.

scripts/add_content_to_db.py
```diff
@@ -70,17 +70,37 @@
     }
 
     with open(args.csv, mode="r", encoding="utf-8") as file:
+        reader = csv.DictReader(file)
+        tags = [val for row in reader for val in row["tags"].strip().split(" ")]
```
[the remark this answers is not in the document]
suzinyou: I think it's safer to assume comma separated.

scripts/add_content_to_db.py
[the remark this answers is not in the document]
lickem22: I definitely think that the upload endpoint for AAQ should handle tags if already in DB. I just wasn't sure of the use case of that specific script. I can add a logic to check for that

scripts/add_content_to_db.py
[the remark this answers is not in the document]
lickem22: Comma separated means  string representation of list right ? Because I can't have a comma separated string in a csv file.

scripts/add_content_to_db.py
```diff
@@ -18,7 +18,7 @@
     "See also the --language option.)",
     usage="""
     python add_content_to_db.py [-h] --csv CSV --domain DOMAIN [--language LANGUAGE]
-
+
```
suzinyou, line 21:
```suggestion

```

Weird that traling whitespace got included? One of the recommit hooks should have caught this

scripts/add_content_to_db.py
```diff
@@ -70,17 +70,44 @@
...
+        tags_db = [val["tag_name"] for val in tags_db.json()]
+        tags = set(tags) - set(tags_db)
+        if len(tags) > 0:
+            print(f"Adding new tags: {tags}")
+        for tag in tags:
+            payload = {
+                "tag_name": tag,
+            }
+            response = requests.post(
+                tags_endpoint, json=payload, headers=headers, verify=False
```
suzinyou, line 87:
```suggestion
                tags_endpoint, json=payload, headers=headers
```
Should we add a comment here to also pass `verify=False` if you are running it locally?

scripts/add_content_to_db.py
```diff
@@ -71,30 +71,37 @@
...
+
             if "tag_id" not in response.json():
-                print(f"Failed to create tag: {tag}")
-                continue
+                raise ValueError("Could not create tag: {tag}.")
             response_json = response.json()
             tags_map[response_json["tag_name"]] = response_json["tag_id"]
-            print(f"Status Code: {response.status_code}, Response: {response.text}")
+        else:
+            print("No new tags to add.")
```
suzinyou, line 96: Should this be after the if statement? I didn't know you can have an else statement on its own.. 

Or put the for loop under the if block?

suzinyou: Right now it prints "Adding new tags {..}", adds new tags, and then prints "No new tags to add".

scripts/add_content_to_db.py
```diff
@@ -70,17 +70,44 @@
     }
 
     with open(args.csv, mode="r", encoding="utf-8") as file:
+        reader = csv.DictReader(file)
+        tags = [val.upper() for row in reader for val in eval(row["tags"])]
```
suzinyou, line 74: Ah ok so this expects the `tags` column values to look like `"['tag1', 'tag2']"`. 

Might make sense if we're getting a dump from some DB, but I think often we're getting data from excel/google sheets, and in google sheets, I'd imagine users would write `"tag1, tag2"` instead? (cc: @Tanmay-97  what do you think?)

image

which ends up with csv file that looks like

```
tags
"tag1,tag2"
"tag2, tag3"
```
In which case we can parse this way instead:
```suggestion
        tags = [val.strip().upper() for row in reader for val in row["tags"].split(",")]
```

scripts/add_content_to_db.py
```diff
@@ -70,17 +70,44 @@
...
+        else:
+            print("No new tags to add.")
+        file.seek(0)
         reader = csv.DictReader(file)
         for row in reader:
             if args.language:
                 language = args.language
             else:
                 language = row["language"]
+            content_tags = [tags_map[val.upper()] for val in eval(row["tags"])]
```
suzinyou, line 104: Related to above, if we assume `"tag1, tag2"` instead of `"['tag1', 'tag2']"`, we can use

```suggestion
            content_tags = [tags_map[val.strip().upper()] for val in row["tags"].split(",")]
```

scripts/add_content_to_db.py
suzinyou: For the import feature, we should be more robust and probably want to handle both cases

suzinyou:
I pushed 1 commit to remove the `language` argument from the script, since we will be removing it!

scripts/add_content_to_db.py
amiraliemami: I'm thinking just handle the natural TAG1,TAG2 usecase (which is what our export button creates too), not the weird list situation

suzinyou (approved):

PR #244, opened 2024-06-13
Add user creation endpoint (only accessible by the first user)

amiraliemami:
Reviewer: @sidravi1 
Estimate: 20mins

---

## Ticket

Fixes: [573](https://idinsight.atlassian.net/browse/AAQ-573)

### Goal

- Add an endpoint so we can easily add users without having to edit the DB or editing the `add_users_to_db.py` script and re-deploying.

### Changes

- Added new endpoint that checks for user_id before proceeding
- Added new pytests for this endpoint
- Moved some stuff around
- Script only adds 1 user to the DB now. More users can be created using the endpoint.

### Future Tasks (optional)

- Add ability to recreate passwords (or let another service, e.g. Fief, handle user management)

## How has this been tested?

- Local dev 
- Docker compose deployment
- Pytests

## Checklist

Fill with `x` for completed. 


core_backend/app/user_tools/routers.py
```diff
@@ -7,15 +7,53 @@
...
+async def create_user(
+    user: UserCreateWithPassword,
+    user_db: Annotated[UserDB, Depends(get_current_user)],
+    asession: AsyncSession = Depends(get_async_session),
+) -> UserRetrieve | None:
+    """
+    Create user endpoint. Can only be used by user with ID 1.
+    """
+
+    if user_db.user_id != 1:
```
sidravi1, line 34: So if for reason `admin` ends up not having a `user_id` of 1, it will fail. I guess that's an ok corner case that we can look out for.

E.g. when we deploy this and run this script, `admin` will not have `user_id` of 1 since the db already has a few users.

core_backend/app/users/schemas.py
```diff
@@ -20,3 +20,14 @@ class UserCreateWithPassword(UserCreate):
     password: str
 
     model_config = ConfigDict(from_attributes=True)
+
+
+class UserRetrieve(BaseModel):
```
sidravi1, line 25: Can you just extend UserCreate? Also, is `user_id` important to share back? I think it should be an internal thing and `username` should be the identifier that people get

core_backend/tests/api/test_z_user_tools.py
sidravi1, line 1: What is the `_z_` in the filename for?

deployment/aws/core_backend/bootstrap.sh
```diff
@@ -42,10 +42,6 @@ export USER1_USERNAME=$(get_secret_value ${SECRET_USER_CREDENTIALS} "user1_usern
 export USER1_PASSWORD=$(get_secret_value ${SECRET_USER_CREDENTIALS} "user1_password" "json")
```
sidravi1, line 42: So in secrets, we need to change USER1 to be Admin?
A future cleanup is to just add it ADMIN_USERNAME everywhere but ok for now.

sidravi1 (approved):
Thanks for the quick PR @amiraliemami!
Looks good. Left some small comments

core_backend/app/user_tools/routers.py
amiraliemami: Basically there's no concept of admin haha it's just whoever was the first user is admin. We can be mindful of that until we implement proper superuser permission handling. Actually, we can probably just have an extra column in the user table that is "admin" which could be a bool, and we check for that instead of user_id. Will think about this more later!

core_backend/app/users/schemas.py
amiraliemami: Hmm, maybe it should just be UserCreate... You're right, we don't need to return the user_id I don't think

core_backend/tests/api/test_z_user_tools.py
amiraliemami: I'm ashamed to say that this is because the tests are not fully isolated and if the key rotation test runs before some other ones it changes user2's API key and so some other tests fail, so it has to run last. I'll refactor this in the future for sure (I think a bunch of other things about tests are also coupled like this)

deployment/aws/core_backend/bootstrap.sh
amiraliemami: Yeah if we want, but doesn't matter what it is really. We should do that in a future cleanup for clarity.

suzinyou (commented):
We should also clean up terraform code that creates user2 credentials

PR #239, opened 2024-06-11
Add content download button

amiraliemami:
Reviewer: @sidravi1 for code, @Tanmay-97 for design.
Estimate: 20mins

---

## Ticket

Fixes: [555](https://idinsight.atlassian.net/browse/AAQ-555) and [572](https://idinsight.atlassian.net/browse/AAQ-572) (new)

### Goal

- Add functionality to download data as a CSV from the front-end

### Changes

- Button and modal and logic for downloading the CSV. Fetches contents and tags and does some cleanup before creating and saving the CSV.
- Some functional and aesthetic cleanups elsewhere in the app

### Future Tasks

- Standardise snackbars across the app. It's getting confusing.

## How has this been tested?

- Run dev and docker-compose deployment, add some contents with tags etc and click download and check the file

## Checklist

Fill with `x` for completed. 

- [x] My code follows the style guidelines of this project
- [x] I have reviewed my own code to ensure good quality
- [x] I have tested the functionality of my code to ensure it works as intended
- [x] I have resolved merge conflicts
- [x] I have updated the requirements

PR #245, opened 2024-06-14
Admin app hotfix

amiraliemami:
Fix admin app build (broken by #239)

### DIFF
--- a/admin_app/src/components/DownloadModal.tsx
+++ b/admin_app/src/components/DownloadModal.tsx
@@ -15,6 +15,19 @@ import { LoadingButton } from "@mui/lab";
 
 const MAX_CARDS_TO_FETCH = 200;
 
+interface Content {
+  user_id: string;
+  content_id: string;
+  content_metadata: Record<string, any>;
+  content_tags: number[];
+  content_tag_names?: string[];
+}
+
+interface Tag {
+  tag_id: string;
+  tag_name: string;
+}
+
 const DownloadModal = ({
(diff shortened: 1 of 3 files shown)


admin_app/src/components/DownloadModal.tsx
```diff
@@ -15,6 +15,19 @@ import { LoadingButton } from "@mui/lab";
 
 const MAX_CARDS_TO_FETCH = 200;
 
+interface Content {
```
sidravi1, line 18: Also defined in "@/app/content/edit/page".
Can we just define it in one place and import it where needed?

sidravi1: Will avoid bugs where we change one and forget to change the other

admin_app/src/components/DownloadModal.tsx
```diff
@@ -15,6 +15,19 @@ import { LoadingButton } from "@mui/lab";
...
+  user_id: string;
+  content_id: string;
+  content_metadata: Record<string, any>;
+  content_tags: number[];
+  content_tag_names?: string[];
+}
+
+interface Tag {
```
sidravi1, line 26: Same here. Also defined in "@/app/content/page".

sidravi1 (commented):
Thanks @amiraliemami - Looks like already merged but for next PR, we shouldn't have the same types/interfaces defined in multiple files. Will cause weird bugs

PR #248, opened 2024-06-17
Fixing dashboard tests

lickem22:
Reviewer: @sidravi1 
Estimate: @20mins

---

## Ticket

Fixes: HOTFIX

### Goal
The goal of this PR is to improve dashboards so that it does not only check for the length of the array returned, but for the content as well 
## How has this been tested?
 Ran the tests

## Checklist

Fill with `x` for completed. 

### DIFF
--- a/core_backend/tests/api/test_question_dashboard.py
+++ b/core_backend/tests/api/test_question_dashboard.py
@@ -1,8 +1,10 @@
-from functools import partial
 import random
-from core_backend.app.question_dashboard.schemas import QuestionDashBoard
+from functools import partial
+
 import pytest
 from fastapi.testclient import TestClient
+
+from core_backend.app.question_dashboard.schemas import QuestionDashBoard
 
 
 async def mock_dashboard_stats(
--- a/core_backend/tests/api/test_question_dashboard.py
+++ b/core_backend/tests/api/test_question_dashboard.py
@@ -1,17 +1,44 @@
+from functools import partial
+import random
+from core_backend.app.question_dashboard.schemas import QuestionDashBoard
 import pytest
 from fastapi.testclient import TestClient
 
-from .conftest import mock_dashboard_stats
+
+async def mock_dashboard_stats(
+    user_id: int, asession: None, questions: tuple, upvotes: tuple
+) -> QuestionDashBoard:
+    """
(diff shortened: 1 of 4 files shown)


core_backend/tests/api/test_question_dashboard.py
```diff
@@ -23,3 +52,7 @@ async def test_retrieve_question_dashboard(
     data = response.json()
     assert len(data["six_months_questions"]) == 6
     assert len(data["six_months_upvotes"]) == 6
+    assert all(value >= questions[0] for value in data["six_months_questions"])
```
lickem22, line 55: Instead of checking each values individually I make sure the values are within the treshold?

core_backend/tests/api/test_question_dashboard.py
```diff
@@ -1,17 +1,46 @@
...
+async def mock_dashboard_stats(
+    user_id: int, asession: None, questions: tuple, upvotes: tuple
+) -> QuestionDashBoard:
+    """
+    Function to monkeypatch question_dashboard.routersaz.get_dashboard_stats with
+    random data
+    """
+    return QuestionDashBoard(
+        six_months_questions=[
+            random.randint(questions[0], questions[1]) for _ in range(6)
```
sidravi1, line 19: My goal was to check for corner cases. I see that you are doing that by including (0, 0) but this may just make it more complex (took me a while to understand it) and I'm unsure of the benefit. 

Can we set a seed to make this deterministic / replicable?

core_backend/tests/api/test_question_dashboard.py
sidravi1: My vote is to just make it fixed values. See comment above!

sidravi1 (approved):
Thanks @lickem22!

Suggest just making the test simpler by using fixed values.

core_backend/tests/api/test_question_dashboard.py
lickem22: I did it that way because it is just too much static code to do it that way, because you need to enter 6 values twice for every case and then check it one by one. Is there any value to make it replicable?

sidravi1:
Just nice to have deterministic tests so if it passes locally we are
confident it will pass in CI/CD. Please add a seed if you want to keep it
the current way

PR #250, opened 2024-06-17
Changed UD to make just one LLM call per request

sidravi1:
Reviewer: @suzinyou 
Estimate: 30mins

---

## Ticket

Fixes: N/A

While writing scripts for testing UD entailment, realised that the number of calls being made were not sustainable (one per id

### Goal

`llm_entailment` method only calls LLM once with all rules in context

### Changes

Updated prompt
Updated `llm_entailment_classifier`
Updated tests

## How has this been tested?

Tests pass

## Checklist

Fill with `x` for completed. 

### DIFF
--- a/core_backend/app/llm_call/entailment.py
+++ b/core_backend/app/llm_call/entailment.py
@@ -17,7 +17,6 @@ async def detect_urgency(
     Detects the urgency of a message based on a set of urgency rules.
     """
 
-    "\n".join([f"{i+1}. {rule}" for i, rule in enumerate(urgency_rules)])
 
     ud_entailment = UrgencyDetectionEntailment(urgency_rules=urgency_rules)
     prompt = ud_entailment.get_prompt()
--- a/core_backend/app/llm_call/entailment.py
+++ b/core_backend/app/llm_call/entailment.py
@@ -17,7 +17,6 @@ async def detect_urgency(
     Detects the urgency of a message based on a set of urgency rules.
     """
 
-
     ud_entailment = UrgencyDetectionEntailment(urgency_rules=urgency_rules)
     prompt = ud_entailment.get_prompt()
 
--- a/core_backend/app/llm_call/entailment.py
+++ b/core_backend/app/llm_call/entailment.py
@@ -1,5 +1,5 @@
 from json import loads
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 from ..config import LITELLM_MODEL_URGENCY_DETECT
 from .llm_prompts import get_urgency_detection_prompt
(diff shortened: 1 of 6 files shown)


core_backend/app/llm_call/llm_prompts.py
```diff
@@ -292,29 +292,100 @@ class AlignmentScore(BaseModel):
...
+        """
+        self._urgency_rules = urgency_rules
+
+    def parse_json(self, json_str: str) -> Dict:
+        """
+        Validates the output of the urgency detection entailment task.
+        """
+
+        json_str = json_str.replace("```json", "").replace("```", "")
+        json_str = json_str.replace("\{", "{").replace("\}", "}")
```
suzinyou, line 355: I created a function `core_backend.app.llm_calls.utils.remove_json_markdown` -- can we update that one and use it here?

core_backend/app/llm_call/llm_prompts.py
```diff
@@ -292,29 +292,100 @@ class AlignmentScore(BaseModel):
...
+
+        # fmt: off
+        json = (
```
suzinyou, line 358: nit: can we name this smth else because it looks like the json library?

core_backend/app/llm_call/llm_prompts.py
```diff
@@ -292,29 +292,100 @@ class AlignmentScore(BaseModel):
...
+            raise ValueError(
+                (
+                    f"Best_matching_rule {json.best_matching_rule} is "
+                    f"not in the urgency rules provided."
+                )
+            )
+
+        return json.model_dump()
+
+    def get_prompt(self) -> str:
```
suzinyou, line 377: Neat! We'll do this with other prompts, too!

core_backend/app/llm_call/entailment.py
```diff
@@ -1,24 +1,38 @@
...
+
+    ud_entailment = UrgencyDetectionEntailment(urgency_rules=urgency_rules)
+    prompt = ud_entailment.get_prompt()
+
+    json_str = await _ask_llm_async(
+        question=message,
         prompt=prompt,
         litellm_model=LITELLM_MODEL_URGENCY_DETECT,
         metadata=metadata,
     )
```
suzinyou, line 30: You can also add `json=true` to optionally use OpenAI's json response format!

core_backend/tests/api/test_urgency_detect.py
```diff
@@ -65,6 +67,7 @@ async def test_user2_access_user1_rules(
 
         if response.status_code == 200:
             is_urgent = response.json()["is_urgent"]
+            print(response.json())
```
suzinyou, line 70:
```suggestion
```

core_backend/app/llm_call/llm_prompts.py
```diff
@@ -292,29 +292,100 @@ class AlignmentScore(BaseModel):
...
+        json = (
+            UrgencyDetectionEntailment
+                .UrgencyDetectionEntailmentResult
+                .model_validate_json(
+                    json_str
+                )
+            )
+        # fmt: on
+
+        if json.best_matching_rule not in self._urgency_rules:
```
suzinyou, line 367: image
The rule was exactly "How many advils" -- should we add a clean up step for this field?

Also, when the error is raised, we return the default response. We should have an error response for urgency detection. And/or at least update the response json with valid json fields maybe as a temporary solution?

```json
{
  "is_urgent": false,
  "failed_rules": [],
  "details": {
    "best_matching_rule": "",
    "probability": 0,
    "reason": ""
  }
}
```

suzinyou (changes requested):
Thanks Sid! This makes sense, but dealing with json gets tricky :/ let me know what you think about my comments

core_backend/app/llm_call/entailment.py
sidravi1: Will it just get ignored if it's an LLM that doesn't use this?

suzinyou: It should get ignored! passing json=true will pass an additional kwarg `response_format`
https://docs.litellm.ai/docs/completion/input

core_backend/app/llm_call/llm_prompts.py
sidravi1: Added a regex to strip the number from it. Thanks for picking this up.

And when it fails, it raises a warning and returns default json - see `detect_urgency` in `core_backend/app/llm_call/entailment.py`. That's actually what is throwing the JSONDecode Error.

PR #254, opened 2024-06-22
CSV content import endpoint (no tags)

amiraliemami:
Reviewer: @lickem22 
Estimate: 30mins

---

## Ticket

Fixes: [596](https://idinsight.atlassian.net/browse/AAQ-596)

### Goal

- This PR adds a new endpoint that accepts a CSV and imports new contents. Note: no support for tags yet.

### Changes

- Adds endpoint
- Adds pandas to requirements
- Adds new tests

### Future Tasks

- Add UI
- Add support for tags

## How has this been tested?
- dev swagger UI
- make tests
- docker compose deployment

## Checklist

- [x ] I have tested the functionality of my code to ensure it works as intended

### DIFF
--- a/core_backend/app/contents/routers.py
+++ b/core_backend/app/contents/routers.py
@@ -165,12 +165,14 @@ async def upload_contents_in_bulk(
     file: UploadFile,
     user_db: Annotated[UserDB, Depends(get_current_user)],
     asession: AsyncSession = Depends(get_async_session),
-) -> List[ContentDB] | None:
+) -> List[ContentRetrieve] | None:
     """
     Upload and process contents in bulk from a CSV file.
     """
 
     # TODO: deal with tags!
+    if file is None:
+        raise HTTPException(status_code=400, detail="No file uploaded")
 
     # Ensure the file is a CSV
     if not file.filename.endswith(".csv"):
--- a/core_backend/app/contents/routers.py
+++ b/core_backend/app/contents/routers.py
@@ -286,6 +286,13 @@ async def _csv_checks(df: pd.DataFrame, user_id: int, asession: AsyncSession) ->
         error_list_model = CustomErrorList(errors=error_list)
         raise HTTPException(status_code=400, detail=error_list_model.dict())
     else:
+        # strip columns to catch duplicates better and empty cells
+        df["content_title"] = df["content_title"].str.strip()
+        df["content_text"] = df["content_text"].str.strip()
+
+        # set any empty strings to None
+        df = df.replace("", None)
+
         # check if there are any empty values in either column
         if df["content_title"].isnull().any():
             error_list.append(
@@ -323,8 +330,6 @@ async def _csv_checks(df: pd.DataFrame, user_id: int, asession: AsyncSession) ->
             )
 
         # check if there are duplicates in either column
-        df["content_title"] = df["content_title"].str.strip()
-        df["content_text"] = df["content_text"].str.strip()
         if df.duplicated(subset=["content_title"]).any():
             error_list.append(
                 CustomError(
--- a/core_backend/app/contents/routers.py
+++ b/core_backend/app/contents/routers.py
@@ -207,7 +207,6 @@ def _load_csv(file: UploadFile) -> pd.DataFrame:
 
     try:
         df = pd.read_csv(file.file)
-        return df
     except EmptyDataError as e:
         raise HTTPException(status_code=400, detail="The CSV file is empty") from e
     except ParserError as e:
@@ -218,6 +217,10 @@ def _load_csv(file: UploadFile) -> pd.DataFrame:
         raise HTTPException(
             status_code=400, detail="CSV is unreadable (encoding error)"
         ) from e
+    if df.empty:
+        raise HTTPException(status_code=400, detail="The CSV file is empty")
+
+    return df
 
 
 async def _csv_checks(df: pd.DataFrame, user_id: int, asession: AsyncSession) -> None:
--- a/core_backend/app/contents/routers.py
+++ b/core_backend/app/contents/routers.py
@@ -161,7 +161,7 @@ async def retrieve_content_by_id(
 
 
 @router.post("/csv-upload", response_model=List[ContentRetrieve])
-async def upload_contents_in_bulk(
+async def bulk_upload_contents(
     file: UploadFile,
     user_db: Annotated[UserDB, Depends(get_current_user)],
     asession: AsyncSession = Depends(get_async_session),
--- a/core_backend/app/contents/routers.py
+++ b/core_backend/app/contents/routers.py
@@ -201,9 +201,9 @@ async def bulk_upload_contents(
         content = ContentCreate(
             content_title=row["content_title"],
             content_text=row["content_text"],
-            content_language=row.get("content_language", "ENGLISH"),
+            content_language="ENGLISH",
             content_tags=[],
-            content_metadata=row.get("content_metadata", {}),
+            content_metadata={},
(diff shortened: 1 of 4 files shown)


core_backend/app/contents/routers.py
```diff
@@ -158,6 +165,209 @@ async def retrieve_content_by_id(
...
+    """
+
+    # TODO: deal with tags!
+
+    # Ensure the file is a CSV
+    if file.filename is None or not file.filename.endswith(".csv"):
+        error_list_model = CustomErrorList(
+            errors=[
+                CustomError(
+                    type="invalid_format",
```
lickem22, line 188: Since we are only dealing with contents, I think we should use `title` instead of `content_title`. It would make it more user friendly and would lead to less errors (For example `content-title` instead of `content_title`). Same for content_text

core_backend/app/contents/routers.py
```diff
@@ -158,6 +165,209 @@ async def retrieve_content_by_id(
...
+                    description="Please upload a CSV file.",
+                )
+            ]
+        )
```
lickem22, line 192: We are not planning on inserting metadata using the upload button?

core_backend/app/contents/routers.py
```diff
@@ -158,6 +165,209 @@ async def retrieve_content_by_id(
...
+def _load_csv(file: UploadFile) -> pd.DataFrame:
+    """
+    Load the CSV file into a pandas DataFrame
+    """
+
+    try:
+        df = pd.read_csv(file.file, dtype=str)
+    except EmptyDataError as e:
+        error_list_model = CustomErrorList(
+            errors=[
```
lickem22, line 226: I think it would be neater if you return a bool value and a string specifying the details of the error instead of instanciating a new HTTPException for each error.
Something list this
```
if "content_title" not in cols or "content_text" not in cols:
        return False, "File must have 'content_title' and 'content_text' columns."
```

core_backend/app/contents/routers.py
```diff
@@ -158,6 +165,209 @@ async def retrieve_content_by_id(
...
+
+async def _csv_checks(df: pd.DataFrame, user_id: int, asession: AsyncSession) -> None:
+    """
+    Perform checks on the CSV file to ensure it meets the requirements
+    """
+
+    # check if content_title and content_text columns are present
+    cols = df.columns
+    error_list = []
+    if "content_title" not in cols or "content_text" not in cols:
```
lickem22, line 276: Maybe strip all texts in the db and csv just to be sure.

core_backend/tests/api/test_import_content.py
```diff
@@ -0,0 +1,265 @@
+from io import BytesIO
+from typing import Generator
+
+import pandas as pd
+import pytest
+from fastapi.testclient import TestClient
+
+
+# to be used for the DB duplicate checks later
+@pytest.fixture(scope="function")
```
lickem22, line 10: Why not use the already existing feature for `existing_content_id`?

core_backend/tests/api/test_import_content.py
```diff
@@ -0,0 +1,265 @@
...
+        "content_title": [],
+        "content_text": [],
+    }
+    return _dict_to_csv_bytes(data)
+
+
+@pytest.fixture
+def data_missing_columns() -> BytesIO:
+    data = {
+        "wrong_column_1": ["Value 1", "Value 2"],
```
lickem22, line 74: You should test titles being only trailing spaces.

core_backend/tests/api/test_import_content.py
```diff
@@ -0,0 +1,265 @@
...
+def data_duplicate_texts() -> BytesIO:
+    data = {
+        "content_title": ["Title 1", "Title 2"],
+        "content_text": ["Duplicate text", "Duplicate text"],
+    }
+    return _dict_to_csv_bytes(data)
+
+
+@pytest.fixture
+def data_title_in_db() -> BytesIO:
```
lickem22, line 135: Maybe test behaviour when there is a mismatch between the  number of titles and the number of texts

core_backend/tests/api/test_import_content.py
```diff
@@ -0,0 +1,265 @@
...
+                headers={"Authorization": f"Bearer {fullaccess_token}"},
+            )
+            assert response.status_code == 200
+
+    @pytest.mark.parametrize(
+        "mock_csv_data, expected_error_type",
+        [
+            (
+                "data_empty_csv",
+                "empty_data",
```
lickem22, line 182: We should add these as a collection to avoid repetition.

core_backend/requirements.txt
```diff
@@ -13,3 +13,6 @@ prometheus_client==0.19.0
 google-api-python-client==2.129.0
 google-api-python-client-stubs==1.25.0
 langfuse==2.27.3
+pandas==2.2.2
+pandas-stubs==2.2.2.240603
+types-openpyxl==3.1.4.20240621
```
lickem22, line 18: Out of curiosity, what is `types-openyxl` for ?

lickem22 (commented):
Thanks Amir. Left a few comments and happy to discuss them

lickem22 (changes requested):
> Thanks Amir. Left a few comments and happy to discuss them

core_backend/app/contents/routers.py
amiraliemami: I chose this as per Tanmay's design and also how our export data looks like, but I do agree. I think in a later PR we should rename our endpoint response, export function, and this import function.

core_backend/app/contents/routers.py
amiraliemami: Not yet, will add this later

core_backend/app/contents/routers.py
amiraliemami: I did it this way so we can get the benefits of Pydantic + keep consistency if we add more checks in the future. Also, these errors are always returned as a list so that we can display them altogether on the front-end if we collect multiple (see _csv_checks). Not 100% the way I've done it is the best though so open to changing - @sidravi1 what do you think?

core_backend/app/contents/routers.py
amiraliemami: ahh for duplicate check? will do

core_backend/tests/api/test_import_content.py
amiraliemami: collection like collect them in a list beforehand? good idea

core_backend/tests/api/test_import_content.py
amiraliemami: That'd result in a parsing error I think - I should probably try loading some broken CSVs to test that...

core_backend/tests/api/test_import_content.py
amiraliemami: done

amiraliemami: This was clever, caught a bug with this test!

core_backend/tests/api/test_import_content.py
amiraliemami: because we need specific values for content_title and content_text here, but that one is parameterised with values specific to that file... This is not the best, I agree.

core_backend/requirements.txt
amiraliemami: something mypy suggested

core_backend/tests/api/test_import_content.py
lickem22: Then I suggest calling it something else to avoid confusion.

PR #262, opened 2024-06-27
[128] - TTS functionality for AAQ

MustafaAkolawala:
Reviewer: @lickem22 
Estimate: 20 mins

---

## Ticket

Fixes:  [Ticket link](https://idinsight.atlassian.net/browse/AAQ-588?atlOrigin=eyJpIjoiNTlmY2M3OThjMTViNGIzNDg1ZmU3NzViMTE0NDI1NDciLCJwIjoiaiJ9)

## Description

This PR introduces text-to-speech (TTS) functionality to AAQ's ` llm-response` endpoint, enhancing user accessibility by allowing responses to be delivered in audio format as well as text. A reusable TTS component, `generate_speech` has been created to facilitate this functionality across different parts of the application.

### Goal

The primary goal of this PR is to increase the accessibility of the AAQ platform. By integrating TTS capabilities, users who prefer auditory information can now listen to responses generated by the language model. This functionality ensures that AAQ is more inclusive and caters to a broader audience.

### Changes

- **New Directory Creation:** A new directory named voice_api has been created to house components related to STT and TTS functionality, ensuring better organization and modularity.

- **Endpoint Modification:** The llm-response endpoint has been modified to include TTS functionality. This involves converting text responses to speech and providing a downloadable MP3 file as part of the response, also the `QueryResponse` pydantic model got modified by adding a `tts_file`  as optional to it.

- **Reusable Component:** A reusable TTS component called generate_speech has been implemented. This component takes text input and generates an MP3 file using the Google Text-to-Speech (gTTS) API.

- **Testing:** Unit tests and integration tests have been created and updated to ensure the new TTS functionality works as intended and does not disrupt existing features.

## How has this been tested?

dev-environment
docker-compose
Pytests

Steps to Test the TTS functionality

1. **Enable TTS Functionality:** Provide a flag in `query_metadata` with `"use_tts": true` to enable TTS functionality. 
2. **Generate TTS Response:** Make a request to the llm-response endpoint with the TTS flag enabled. The TTS file will be generated and stored in the ` core_backend` Docker container.
3. **Retrieve the TTS File:** Use the following command in your terminal to copy the file from the Docker container to your local repository root:
`docker cp <container_name/id>:/usr/src/aaq_backend/<name.mp3> ./<name.mp3>`
 Replace <container_name/id> with the appropriate container name or ID, and <name.mp3> with the name of the generated MP3 file.
4. **Verify the TTS File:** Once the file is copied to your local repository, you can play the MP3 file to verify the TTS functionality.

## Checklist

Fill with `x` for completed. 


core_backend/app/question_answer/routers.py
```diff
@@ -140,6 +142,24 @@ async def get_llm_answer(
 
         response.debug_info["extracted_info"] = rag_response.extracted_info
 
+        tts_save_path = f"response_{response.query_id}.mp3"
+
+        if question.query_metadata.get("use_tts", False):
```
lickem22, line 147: We should pass `use_tts` in the schema and not inside the metadata

MustafaAkolawala: so in the `QueryBase` schema just add another optional flag right for `use_tts`?

core_backend/tests/api/test_question_answer.py
```diff
@@ -250,6 +251,14 @@ async def test_content_feedback_check_content_id(
         assert response.status_code == response_code
```
lickem22, line 253: Since you are using it twice, put it in `conftest.py`

core_backend/tests/api/test_question_answer.py
```diff
@@ -314,6 +323,51 @@ async def test_user2_access_user1_content(
...
+                "/llm-response",
+                headers={"Authorization": f"Bearer {token}"},
+                json=user_query,
+            )
+
+            assert response.status_code == expected_status_code
+
+            if expected_status_code == 200:
+                json_response = response.json()
+                print(f"JSON response: {json_response}")
```
lickem22, line 359: You shouldn't print the JSON response

lickem22 (changes requested):
Thanks for this Mustadfa. I left a few comments but main things to change are:
- Put STT logic within `question_answer`
- Refactor a few tests
- I try testing STT locally but I am getting this:
```
"state": "error",
  "tts_file": "response_1.mp3"
```
Is there something I am missing? Like an api key to use gTTS ? 
I am happy to jump on a call to discuss that.

core_backend/tests/api/test_question_answer.py
```diff
@@ -314,6 +323,51 @@ async def test_user2_access_user1_content(
...
+                if "tts_file" in json_response:
+                    if expect_tts_file:
+                        assert json_response["tts_file"] is not None
+                    else:
+                        assert json_response["tts_file"] is None
+                else:
+                    assert (
+                        not expect_tts_file
+                    ), "Expected 'tts_file' but it was not found"
+
```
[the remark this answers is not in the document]
MustafaAkolawala:
```
if expect_tts_file:
    assert "tts_file" in json_response
    assert json_response["tts_file"] is not None
 else:
     assert "tts_file" not in json_response
```

does something like this makes sense?

core_backend/app/question_answer/routers.py
lickem22: Yes

core_backend/tests/api/test_question_answer.py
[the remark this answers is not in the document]
lickem22: Yes that works!

core_backend/app/question_answer/models.py
```diff
@@ -131,6 +134,11 @@ async def save_query_response_to_db(
         llm_response=response.model_dump()["llm_response"],
         response_datetime_utc=datetime.utcnow(),
     )
+
```
lickem22, line 137: You don't need to do that since you gave a default value to `tts_file` in the pydantic model.

core_backend/tests/api/test_TTS.py
```diff
@@ -0,0 +1,52 @@
...
+        self, language: IdentifiedLanguage, text: str, mock_gtts: MagicMock
+    ) -> None:
+        save_path = "test_audio.mp3"
+
+        with patch("core_backend.app.voice_api.text_to_speech.gTTS", mock_gtts):
+            tts_file_path = await generate_speech(text, language, save_path)
+            assert tts_file_path == save_path
+
+    @pytest.mark.parametrize(
+        "text, language",
```
lickem22, line 37: You should add another use case when you type gibberish and the language is supported. Also I think it's worth checking when the text is in a language (like hindi) and the specified language is different (for ex Swahili) to see the behaviour.

core_backend/app/question_answer/routers.py
```diff
@@ -140,6 +142,24 @@ async def get_llm_answer(
 
         response.debug_info["extracted_info"] = rag_response.extracted_info
 
+        tts_save_path = f"response_{response.query_id}.mp3"
```
lickem22, line 145: I just thought about something, in the audio support flow, the audio we are supposed to return is not the question asked by the user but the answer by the LLM, so I suggest we don't modify the question answer module for now but later after we have the full audio support logic. Sorry for not flagging this earlier. There are two ways you can go about that since you are not merging to main:
- Deleting every code you updated in `question_answer` to update it later when needed
- only delete this section for now and return None for the `tts_file`until you implement that logic. 
I leave it up to you how you want to do it. But in both cases, you should remove the changes you made in `test_question_answer`.

lickem22 (approved):
Thanks Mustafa, 
I only have one comment about returning the question as an audio in the response. I will leave it up to you how to update it. 
Also, it is probably not. something to worry about now, but how do we handle errors when the TTS function returns an audio but the only thing said in the audio is `error`. I thought in case of an error, an exception will be raised. I will approve this PR but please implement the changes I made before merging.

core_backend/tests/api/test_TTS.py
MustafaAkolawala: it would not raise any error right now for these specific test cases cause: 
1) as the TTS is being used fairly late in the workflow (after we get a response from the LLM) the guardrails would already catch errors and not let the TTS file get generated at all.
2) as the external TTS we using right now is GTTS, it is limited in it capabilities and cannot identify if the langauge would be gibberish or there is a language mismatch.

but i think that should not be an issue right as when used with the rest of the workflow and with the necessary guardrails implemented (as we would mostly get the correct language identified and it wouldnt give an answer to gibberish) these wouldnt pose a risk, but we can still try implementing extra error handling in the inhouse TTS

lickem22: It makes sense not to test TTS doing that on question_answer, but here we are testing the TTS function alone, so We should make sure the function is robust enough to catch these issues.

MustafaAkolawala: so what do u suggest, right now the external API is not robust enough to catch these errors but when we replace it with an actual neural model we could catch these errors at that time so for now should i merge it?

lickem22: If the api don't catch it then the test wont pass. You can merge for now

PR #268, opened 2024-06-28
Dashboard/stats cards2

sidravi1:
Reviewer:
Estimate:

---

## Ticket

Fixes: JIRA_TICKET_LINK

## Checklist

Fill with `x` for completed. 


core_backend/app/dashboard/models.py
```diff
@@ -0,0 +1,256 @@
...
+        urgency_stats=urgency_stats,
+    )
+
+
+async def get_query_count_stats(
+    user_id: int, asession: AsyncSession, start_date: date, end_date: date
+) -> QueryStats:
+    """
+    Retrieve statistics for question answering
+    """
```
lickem22, line 52: Would it be possible to get that in a single query?

lickem22: Actually, since we will need a line chart as well, maybe just have an endpoint you will use for both?

core_backend/app/dashboard/models.py
```diff
@@ -0,0 +1,256 @@
...
+) -> Dict[str, Union[int, float]]:
+    """
+    Get feedback statistics
+    """
+    n_positive_curr = feedback_curr_period_dict.get("positive", 0)
+    n_negative_curr = feedback_curr_period_dict.get("negative", 0)
+    n_positive_prev = feedback_prev_period_dict.get("positive", 0)
+    n_negative_prev = feedback_prev_period_dict.get("negative", 0)
+
+    if n_positive_prev == 0:
```
lickem22, line 190: I think we should have a function like:
```
def get start(curr,prev):
     if prev == 0:
        percentage_positive_increase = 0.0
    else:
        percentage_positive_increase = (
            curr - prev
        ) / prev
    return percentage_positive_increase

```
So that we can use it twice in that function and in other functions using that logic

admin_app/src/app/dashboard/components/Overview.tsx
```diff
@@ -1,7 +1,95 @@
...
+  useEffect(() => {
+    getStatsCardData(timePeriod, token!).then((data) => {
+      const {
+        content_feedback_stats,
+        query_stats,
+        response_feedback_stats,
+        urgency_stats,
+      } = data.stats_cards;
+
+      console.log(data.stats_cards);
```
lickem22, line 30: We should delete the log here.

admin_app/src/app/dashboard/components/Overview.tsx
```diff
@@ -1,7 +1,95 @@
...
+
+      parseCardData(data, timePeriod);
+      // Total Queries
```
lickem22, line 33: Seems like this comment should be deleted as well

admin_app/src/app/dashboard/components/StatCard.tsx
```diff
@@ -0,0 +1,121 @@
+import React from "react";
+import { Box, Typography } from "@mui/material";
+import ArrowDownwardIcon from "@mui/icons-material/ArrowDownward";
+import HeadsetMicIcon from "@mui/icons-material/HeadsetMic";
```
lickem22, line 4: Unused imports

admin_app/src/app/dashboard/api.ts
```diff
@@ -0,0 +1,23 @@
+const NEXT_PUBLIC_BACKEND_URL: string =
+  process.env.NEXT_PUBLIC_BACKEND_URL || "http://localhost:8000";
+
+type Period = "day" | "week" | "month" | "year";
```
lickem22, line 4: Can you export the type from `types.ts` so you don't have to rewrite the code?

admin_app/src/app/dashboard/page.tsx
```diff
@@ -4,22 +4,24 @@ import React from "react";
 import { Box, Container, Grid, Typography } from "@mui/material";
```
lickem22, line 4: Unused imports

admin_app/src/app/dashboard/page.tsx
```diff
@@ -4,22 +4,24 @@ import React from "react";
...
 import { Sidebar, PageName } from "@/app/dashboard/components/Sidebar";
 import TabPanel from "@/app/dashboard/components/TabPanel";
+import { Period } from "./types";
 import Overview from "@/app/dashboard/components/Overview";
 import { useState } from "react";
 import { Global, css } from "@emotion/react";
+import { useEffect } from "react";
```
lickem22, line 11: Unused imports

admin_app/src/app/dashboard/components/Overview.tsx
```diff
@@ -1,7 +1,95 @@
...
+import { useEffect } from "react";
+
+interface OverviewProps {
+  timePeriod: Period;
+}
+
+const Overview: React.FC<OverviewProps> = ({ timePeriod }) => {
+  const { token } = useAuth();
+  const [statCardData, setStatCardData] = React.useState<StatCardProps[]>([]);
+  useEffect(() => {
```
lickem22, line 21: I remember you mentioned components should be dull and thus should not call APIs, should we do the same for this and make the api call inside the page rather than the component?

admin_app/src/app/dashboard/components/TabPanel.tsx
```diff
@@ -26,3 +37,4 @@ const TabPanel: React.FC<TabPanelProps> = ({ tabValue, handleChange }) => {
 };
 
 export default TabPanel;
+export type { Period };
```
lickem22, line 40: What's the use for this ?

admin_app/src/app/dashboard/types.ts
```diff
@@ -0,0 +1,4 @@
+type Period = "day" | "week" | "month" | "year";
+type TimeFrame = "Last 24 hours" | "Last week" | "Last month" | "Last year";
+
+export type { Period, TimeFrame };
```
lickem22, line 4: Maybe we should call this `utils.ts` to keep consistency accross modules going forward. Then we will be able to add `tabLabels` from `TabPanel.tsx`.

core_backend/tests/api/test_dashboard.py
```diff
@@ -0,0 +1,204 @@
...
+                is_urgent=is_urgent, failed_rules=[], details={}
+            )
+            await save_urgency_response_to_db(
+                urgency_query_db, urgency_response, asession
+            )
+
+            urgency_query_ids.append(urgency_query_db.urgency_query_id)
+            urgency_response_ids.append(urgency_query_db.urgency_query_id)
+
+        yield (n_urgent, n_not_urgent)
```
lickem22, line 69: I think we should not yield `n_not_urgent` if we are not using it in the code logic.

core_backend/tests/api/conftest.py
```diff
@@ -151,8 +134,20 @@ async def faq_contents(client: TestClient, db_session: Session) -> None:
         )
         contents.append(content_db)
 
-    db_session.add_all(contents)
-    db_session.commit()
+    asession.add_all(contents)
+    await asession.commit()
+    # db_session.add_all(contents)
+    # db_session.commit()
```
lickem22, line 140: Comments to be deleted

core_backend/tests/api/conftest.py
```diff
@@ -124,14 +102,20 @@ def user(client: TestClient, db_session: Session) -> None:
     # update the TEST_USER_ID global variable
     TEST_USER_ID = user1_db.user_id
 
+    yield user1_db.user_id
 
-@pytest.fixture(scope="session")
-async def faq_contents(client: TestClient, db_session: Session) -> None:
+    # db_session.delete(user1_db)
```
lickem22, line 107: Why are we keeping this as commented?

lickem22 (changes requested):
Thanks for this @sidravi1, I left a few comments. The most important being that I think we need to find a way to make one call to the DB to get both current stats and previous stats. That way, we can limit the number of calls we make to the DB. What do you think?

admin_app/src/app/dashboard/api.ts
sidravi1: Good point! i'll fix this

admin_app/src/app/dashboard/components/Overview.tsx
sidravi1: Hmm yeah. That is true for low level components. Let me think about what to do about higher level ones. This is the entire page 1.

admin_app/src/app/dashboard/components/Overview.tsx
sidravi1: Thanks! Will do

admin_app/src/app/dashboard/components/Overview.tsx
sidravi1: Thanks! will do

admin_app/src/app/dashboard/components/StatCard.tsx
sidravi1: Thanks!

admin_app/src/app/dashboard/components/TabPanel.tsx
sidravi1: I should be importing these from types as you suggested above. I'll fix this.

admin_app/src/app/dashboard/types.ts
sidravi1: I want to keep types together and utils together. I'll have a think about this

core_backend/app/dashboard/models.py
sidravi1: I'll refactor as I add more charts. Though an extra Db call is not that much of an overhead since it's all on the server side. And it makes for nice modularised code

core_backend/app/dashboard/models.py
sidravi1: Good idea. I'll refactor this

core_backend/tests/api/conftest.py
sidravi1: Good point. Will delete.

core_backend/app/dashboard/models.py
sidravi1: Actually, good point. We probably can combine it here since it's all inside one function. Let me do that

core_backend/tests/api/test_dashboard.py
sidravi1: Ah, I am doing this for completeness. I'll update the test to just map this to `_`

core_backend/app/dashboard/models.py
sidravi1: Fixed

PR #275, opened 2024-07-05
[AAQ-558] Implements api calls limits using Redis

lickem22:
Reviewer: @sidravi1 
Estimate: 1h

---

## Ticket

Fixes:[AAQ-558](https://idinsight.atlassian.net/browse/AAQ-558)

### Goal
The goal of this PR is to implement an api calls daily limits on question-answering and urgency-detection. 
### Changes
- Added container for redis
- Added dependency to limit api calls 
- Updated DB to add an api daily quotas
- Added scheduled task to reset daily quotas after 24h
- Updated tests to create users with endpoints
- Added api daily quotas tests
### Future Tasks (optional)
- Implement content quotas using redis
## How has this been tested?
- Tested using dev setup (make setup-redis)
- Tested using deployment (docker compose)
- Tested using swagger UI
## Checklist

Fill with `x` for completed. 

- [x] I have updated the CI/CD scripts in `.github/workflows/`

[AAQ-558]: https://idinsight.atlassian.net/browse/AAQ-558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core_backend/app/__init__.py
```diff
@@ -73,9 +77,37 @@ def create_app() -> FastAPI:
...
+
     @app.on_event("startup")
-    def startup_event() -> None:
+    async def startup_event() -> None:
         """Startup event"""
+        app.state.redis = await aioredis.from_url(redis_host)
         logger.info("Application started")
+        scheduler = AsyncIOScheduler()
+        await reset_quota()
+        scheduler.add_job(reset_quota, "interval", minutes=1440)
```
suzinyou, line 103: Does this mean it resets at 00:00:00 UTC?

core_backend/app/__init__.py
```diff
@@ -73,9 +77,37 @@ def create_app() -> FastAPI:
...
+                        f"remaining-calls:{user.username}",
+                        api_daily_quota,
+                    )
+                logger.info("Successfully Synced Redis to database")
+            except Exception as e:
+                logger.info(f"Redis sync failed: {e}")
+
     @app.on_event("startup")
-    def startup_event() -> None:
+    async def startup_event() -> None:
```
suzinyou, line 97: If we redeploy the backend app in the middle of the day, will the API limits reset?

Makefile
```diff
@@ -51,6 +51,15 @@ setup-db:
 	cd core_backend && \
 	python -m alembic upgrade head
 
+setup-redis:
```
sidravi1, line 54: Can we also include a `teardown-redis`

core_backend/tests/api/conftest.py
```diff
@@ -99,34 +106,58 @@ async def asession(
...
+def user1(client: TestClient, db_session: Session) -> None:
+    stmt = select(UserDB).where(UserDB.username == TEST_USERNAME)
+    result = db_session.execute(stmt)
+    user = result.scalar_one()
+    yield user.user_id
+
+
+@pytest.fixture(scope="session", autouse=True)
+def user(
+    client: TestClient, db_session: Session, admin_user, fullaccess_token_admin: str
```
sidravi1, line 135: Why does it use admin_user as a fixture? Do we need an Admin account to create these two?

deployment/docker-compose/docker-compose.dev.yml
```diff
@@ -13,6 +14,12 @@ services:
       - db_volume:/var/lib/postgresql/data
     ports:
       - 5434:5432
+  redis:
```
sidravi1, line 17: Why do we add this to `docker-compose.dev` and not to `docker-compose`. We'll need it all instances right?

core_backend/app/__init__.py
sidravi1: Should this be based on a time / calendar trigger instead of interval?

sidravi1: Also, ideally it would be X calls in a 24 hrs period. So some sort of rolling limit. But I can see that might be tricky so ok to park it if it's not simple.

core_backend/app/utils.py
```diff
@@ -204,3 +204,11 @@ def get_http_client() -> aiohttp.ClientSession:
         http_client.start()
         _HTTP_CLIENT = http_client()
     return _HTTP_CLIENT
+
+
+def encode_api_limit(api_limit: int | None) -> int | str:
```
sidravi1, line 209: /nit Is this used just once? We could just move this line to the function calling it. Maybe make it easier to understand

core_backend/tests/api/test_question_answer.py
```diff
@@ -22,30 +22,158 @@
...
 )
 from core_backend.tests.api.conftest import (
-    TEST_USER_API_KEY,
-    TEST_USER_API_KEY_2,
+    TEST_USERNAME,
+    TEST_USERNAME_2,
 )
 
 
+class TestApiCallQuota:
```
sidravi1, line 30: Thanks! Can you also test reseting quota?

core_backend/app/user_tools/routers.py
```diff
@@ -42,9 +43,14 @@ async def create_user(
...
             asession=asession,
         )
+        await request.app.state.redis.set(
+            f"remaining-calls:{user_db.username}",
+            encode_api_limit(user_db.api_daily_quota),
+        )
         return UserCreate(
             username=user_db.username,
             content_quota=user_db.content_quota,
+            api_daily_quota=user_db.api_daily_quota,
```
sidravi1, line 53: Why do we copy the quota limit from the user that is creating it? Shouldn't we do it from config?

Also, see `authenticate_or_create_google_user` - we also create an user when someone logs in for the first time using their google email. That is the primary use case where we need limits.

sidravi1 (changes requested):
Looks good overall @lickem22!

Would you mind rebasing and addressing the comments? I'll review it again tomorrow.

Major questions for me are:
- What happens when a worker out of a pool of 5 restarts 
- Should resets based on time (in UTC) instead of intervals. 
- Can we add a test that checks if a user created through google login has the right limit. That's the main use case for this.

core_backend/tests/api/conftest.py
lickem22: Yes, Because I am creating the users using the API and not using the db sesion. And since only the first user (user with user_id=1) can create users I need that users for user creation.

core_backend/app/utils.py
lickem22: It is used twice

core_backend/app/__init__.py
lickem22: I updated the code to make sure we execute it once. By setting a redis value that expires at midnight. So  we only run the job when it expires

core_backend/app/user_tools/routers.py
lickem22: It's not the user that is creating it but the user that is newly created. I see how it is confusing since we are using the same variable name. I will update it.

.github/workflows/deploy_gcp_redis.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+    permissions:
+      contents: "read"
+      id-token: "write"
+
+    # TODO: replace improve-gcp-deploy with main
+    environment: gcp-${{ (github.ref_name == 'main' && 'testing') || github.ref_name }}
+
+    env:
+      RESOURCE_PREFIX: ${{ secrets.PROJECT_NAME }}-${{ (github.ref_name == 'main' && 'testing') || github.ref_name }}
```
suzinyou, line 25: Just a note that this code + trigger conditions will change in #327  -- if this branch gets merged first, can you ping me to update #327 ? If #327 gets merged first, would you mind updating this script (copy-paste!) before merging?

(cc: @markbotterill  also pls let me know when you review/merge #327 !)

.github/workflows/deploy_gcp_redis.yaml
```diff
@@ -0,0 +1,68 @@
...
+          workload_identity_provider: projects/${{ secrets.GCP_PROJECT_NUMBER }}/locations/global/workloadIdentityPools/${{ vars.POOL_ID }}/providers/${{ vars.PROVIDER_ID }}
+          service_account: ${{ secrets.GCP_SERVICE_ACCOUNT_EMAIL }}
+
+      - name: Retrieve secrets from Secret Manager
+        id: "secrets"
+        uses: "google-github-actions/get-secretmanager-secrets@v2"
+        with:
+          secrets: |-
+            domain:${{ secrets.GCP_PROJECT_ID }}/${{ env.RESOURCE_PREFIX }}-domain
+
```
suzinyou, line 44: This is not needed

.github/workflows/deploy_gcp_redis.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Show deployment command output
+        run: |-
+          echo '${{ steps.compute-ssh.outputs.stdout }}'
+          echo '${{ steps.compute-ssh.outputs.stderr }}'
+
+      - name: Wait for Application to start
+        id: wait-for-app
+        run: sleep 1m
+        shell: bash
```
suzinyou, line 68: Is there a way to check that redis is up an running? Idea was to wait for 1 min for container to start (if it takes that long) and then ping some healthcheck endpoint. For example, in the backend's healthcheck endpoint maybe we can check that the backend container connected to redis successfully? (like we do for DB)

If you think that will take too long to develop, let's put it in the backlog and remove this step altogether.

.github/workflows/deploy_gcp_redis.yaml
```diff
@@ -0,0 +1,68 @@
...
+        with:
+          instance_name: "${{ secrets.DEPLOYMENT_INSTANCE_NAME }}"
+          zone: "${{ secrets.DEPLOYMENT_ZONE }}"
+          ssh_private_key: "${{ secrets.GCP_SSH_PRIVATE_KEY }}"
+          command: |
+            docker stop redis
+            docker rm redis
+            docker run --name redis \
+              -p 6379:6379 \
+              -d redis:6.0-alpine
```
suzinyou, line 57: We need to set the GCP log driver and use the same docker network as other containers.
 Not sure if this needs the restart option, though.. This is from the backend script:

```
docker run -d \
              --log-driver=gcplogs \
              --restart always \
              --network aaq-network \
```

core_backend/app/auth/dependencies.py
```diff
@@ -140,3 +149,27 @@ def create_access_token(username: str) -> str:
...
+) -> None:
+    """
+    Rate limiter for the API calls. Gets daily quota and decrement it
+    """
+    username = user_db.username
+    key = f"remaining-calls:{username}"
+    redis = request.app.state.redis
+    ttl = await redis.ttl(key)
+    # if key does not exist, set the key and value
+    if ttl == -2:
```
sidravi1, line 166: So -2 means key doesn't exist? Mind using a var to make this obvious? E.g. KEY_DOES_NOT_EXIST = -2

core_backend/app/auth/routers.py
```diff
@@ -2,6 +2,7 @@
 from fastapi.security import OAuth2PasswordRequestForm
 from google.auth.transport import requests
 from google.oauth2 import id_token
+from starlette.requests import Request
```
sidravi1, line 5: \nit Probably the same but mind using?
```
from fastapi.requests import Request
```

core_backend/app/config.py
```diff
@@ -53,6 +53,7 @@
 )
 CHECK_CONTENT_LIMIT = os.environ.get("CHECK_CONTENT_LIMIT", True)
 DEFAULT_CONTENT_QUOTA = int(os.environ.get("DEFAULT_CONTENT_QUOTA", 50))
+DEFAULT_API_QUOTA = int(os.environ.get("DEFAULT_CONTENT_QUOTA", 100))
```
sidravi1, line 56:
```suggestion
DEFAULT_API_QUOTA = int(os.environ.get("DEFAULT_API_QUOTA", 100))
```

core_backend/app/question_answer/config.py
```diff
@@ -1,4 +1,5 @@
 import os
 
 # Functionality variables
+API_CALLS_LIMIT = os.environ.get("API_CALLS_LIMIT", 20)
```
sidravi1, line 4: How is this different from `DEFAULT_API_QUOTA` in app/config.py ?

core_backend/tests/api/conftest.py
```diff
@@ -80,38 +83,71 @@ async def asession(
...
+        api_daily_quota=None,
+        created_datetime_utc=datetime.utcnow(),
+        updated_datetime_utc=datetime.utcnow(),
     )
-    db_session.add(user1_db)
-    db_session.add(user2_db)
+
+    db_session.add(admin_user)
     db_session.commit()
+    yield admin_user.user_id
```
sidravi1, line 99: We yield `user_id` for the first one but the second one is lost if we don't save it in a global right?

sidravi1: Never mind! I see that you have new fixtures now

sidravi1 (approved):
Wow! Lots of files touched here. Some minor comments but looks good!
Thank you

core_backend/app/auth/routers.py
lickem22: I am not sure if it's the same but I used it because it is the same used in `core_backend.app.prometheus_middleware.py`should I also change it there ?

PR #289, opened 2024-07-15
[Bug fix] make API calls to load tags only if logged in

suzinyou:
Reviewer: @sidravi1 
Estimate: 15 minutes

---

## Ticket

Fixes: -

## Description

When you don't have any tags and open the contents page, we get a react error complaining about failing to fetch tags.

https://github.com/user-attachments/assets/35dee623-cbb8-47fc-b999-4d44a0f40df6

### Changes

Check for token before fetching tags

## Checklist

Fill with `x` for completed. 

### DIFF
--- a/admin_app/src/app/content/page.tsx
+++ b/admin_app/src/app/content/page.tsx
@@ -62,7 +62,7 @@ const CardsPage = () => {
     };
     fetchTags();
     setCurrAccessLevel(accessLevel);
-  }, [accessLevel]);
+  }, [accessLevel, token]);
 
   return (
     <>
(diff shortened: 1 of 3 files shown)


admin_app/src/app/content/page.tsx
```diff
@@ -54,8 +54,10 @@ const CardsPage = () => {
 
   React.useEffect(() => {
     const fetchTags = async () => {
-      const data = await apiCalls.getTagList(token!);
-      setTags(data);
+      if (token) {
```
sidravi1, line 57: Should we add `token` to line 64 so this runs once the token has been fetched?

sidravi1 (approved):
Thanks @suzinyou! One minor comment.

amiraliemami:
Something similar to this also happens on the dashboard page btw

admin_app/src/app/content/page.tsx
sidravi1: I think we need both! But please do try if it works with just adding to line 64

suzinyou: You are right we need both.

suzinyou:
For all the pages that fetch some kindof content based on token, we were actually getting API errors when logging out ("Failed to fetch <urgency rules, dashboard content, ..>" -- in console).

This is because 
1. `token` which is a dependency for these useEffect hooks was changing to `null`
2.  Which triggers the hooks to load content, but the API calls were being made with empty tokens!!!

So I have added code so that, when the hook gets triggered by `token` changing,
1. if token is not null, fetch data
2. if it's null, set data to empty states (`setContent([]);` etc.)

Let me know what you think @sidravi1 @amiraliemami

admin_app/src/app/content/page.tsx
```diff
@@ -316,34 +320,45 @@ const CardsGrid = ({
...
+            message: `Failed to fetch content`,
+            color: "error",
+          });
+          setIsLoading(false);
+        });
+    } else {
+      setCards([]);
+      setMaxPages(1);
+      setIsLoading(false);
+    }
```
suzinyou, line 361: I figured it is safer to set the data to null, even though shortly after we get redirected to login page. Wdyt?

suzinyou:
rebased to main

admin_app/src/app/content/page.tsx
sidravi1: Im ok with it. Thanks!

admin_app/src/app/dashboard/components/Overview.tsx
```diff
@@ -82,12 +82,19 @@ const Overview: React.FC<OverviewProps> = ({ timePeriod }) => {
...
 
   useEffect(() => {
-    getOverviewPageData(timePeriod, token!).then((data) => {
-      parseCardData(data.stats_cards, timePeriod);
-      parseHeatmapData(data.heatmap);
-      parseTimeseriesData(data.time_series);
-      setTopContentData(data.top_content);
-    });
+    if (token) {
+      getOverviewPageData(timePeriod, token!).then((data) => {
```
sidravi1, line 86: Can probably remove that ! after token?

sidravi1 (approved):
lgtm! Thanks @suzinyou

PR #309, opened 2024-07-19
Pylint, doc strings, global variable handling, and tightening up logic.

tonyzhao6:
Reviewer: @sidravi1 
Estimate: ~15 minutes - 30 minutes

---

## Ticket

Fixes: N/A

### Goal

This PR introduces updates doc strings and uses `pylint` as a tool to help tighten up logic.

No new functionalities/features are added. Mostly just refactoring a bit.

### Changes

The majority of the changes are either stylistic, refactoring, or improves the logic.

## How has this been tested?

1. `pylint` was executed manually in local environment.

## Checklist

Fill with `x` for completed. 

- [X] I have updated the requirements

### DIFF
--- a/core_backend/app/admin/routers.py
+++ b/core_backend/app/admin/routers.py
@@ -1,5 +1,6 @@
 from urllib.parse import urlparse
 
+import aiohttp
 from aiohttp.client_exceptions import ClientConnectorError
 from fastapi import APIRouter, Depends
 from fastapi.responses import JSONResponse
@@ -32,6 +33,7 @@ async def healthcheck(
         url = urlparse(ALIGN_SCORE_API)
         healthcheck_url = f"{url.scheme!r}://{url.netloc!r}/healthcheck"
         http_client = get_http_client()
+        assert isinstance(http_client, aiohttp.ClientSession)
         try:
             resp = await http_client.get(healthcheck_url)
         except ClientConnectorError as e:
--- a/core_backend/app/contents/models.py
+++ b/core_backend/app/contents/models.py
@@ -1,3 +1,7 @@
+"""This module contains the ORM for managing content in the `ContentDB` database and
+database helper functions such as saving, updating, deleting, and retrieving content.
+"""
+
 from datetime import datetime
 from typing import Dict, List, Optional
 
@@ -32,8 +36,10 @@ from .schemas import (
 
 
 class ContentDB(Base):
-    """
-    SQL Alchemy data model for content
+    """ORM for managing content.
+
+    This database ties into the Admin app and allows the user to view, add, edit,
+    and delete content in the `content` table.
     """
 
     __tablename__ = "content"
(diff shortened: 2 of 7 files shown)


core_backend/app/utils.py
```diff
@@ -194,13 +197,34 @@ def __call__(self) -> aiohttp.ClientSession:
...
+
+    :param http_client: The value to set for the global variable _HTTP_CLIENT.
     """
+
     global _HTTP_CLIENT
-    if _HTTP_CLIENT is None or _HTTP_CLIENT.closed:
+    _HTTP_CLIENT = http_client()
+
+
+def get_http_client() -> Optional[aiohttp.ClientSession]:
```
sidravi1, line 220: This should not be optional right? Should always return ClientSession?

core_backend/app/admin/routers.py
```diff
@@ -32,6 +33,7 @@ async def healthcheck(
         url = urlparse(ALIGN_SCORE_API)
         healthcheck_url = f"{url.scheme!r}://{url.netloc!r}/healthcheck"
         http_client = get_http_client()
+        assert isinstance(http_client, aiohttp.ClientSession)
```
sidravi1, line 36: Is this needed if we ensure get_http_client always returns ClientSession?

core_backend/app/contents/models.py
```diff
@@ -98,9 +109,16 @@ async def save_content_to_db(
     content: ContentCreate,
     asession: AsyncSession,
 ) -> ContentDB:
+    """Vectorize the content and save to the database.
+
+    :param user_id: The ID of the user requesting the save.
```
sidravi1, line 114: Can we please use numpy style docstrings? That's the standard we use.
Also, we used to be very particular about docstrings but since we started using typing, have found less use for it. Having said that, please add wherever you find the code to not be intuitive

core_backend/app/contents/models.py
```diff
@@ -193,9 +222,18 @@ async def delete_content_from_db(
     content_id: int,
     asession: AsyncSession,
 ) -> None:
+    """Delete content from the database.
+
+    NB: Deletion is a soft delete, meaning that the content is not actually removed.
```
sidravi1, line 227: Remove from here?

sidravi1 (commented):
Some quick comments! Thanks for all these improvements :)

core_backend/app/utils.py
tonyzhao6: this is `Optional` because [this line](https://github.com/IDinsight/aaq-core/blob/351aeb398aa64918bb99a1bb21c0d1fc18a4457e/core_backend/app/utils.py#L194) defines it as a union of two types.

core_backend/app/contents/models.py
tonyzhao6: oh yes, these should be removed!

core_backend/app/contents/models.py
tonyzhao6: 1. can definitely switch to numpy style docstrings.
2. for the most part, the docstrings would not be useful for active/main developers. it would be for users, new developers, and long-term maintenance.

core_backend/app/admin/routers.py
tonyzhao6: i believe so---i'm not sure how the original `def get_http_client()` passed `mypy` checks but without the assertions it is failing pre-commit for me.

sidravi1 (approved):
lgtm! Thanks @tonyzhao6

PR #325, opened 2024-07-25
Return search results even if guardrails fail

amiraliemami:
Reviewer: @sidravi1 
Estimate: 1hr

---

## Ticket

Fixes: [551](https://idinsight.atlassian.net/browse/AAQ-551)

### Goal

- Return search results even if guardrails fail. This allows the chat manager to say smt like "We can't answer that question directly, but we found these references that might help: ..."

### Changes

- Update `/search` flow:
  - Translate and paraphrase run and result is used for embeddings search, but original message is used for generating the LLM response.
  - Add original_language value into LLM response prompt explicitly

- Update schemas for search endpoint to standardise and always return search results even if we send a 400 error.  Update examples to use Pydantic Field.

- Standardise how decorators function:
  - Remove `state` variable and the use of kwargs. Everything is now passed within the `QueryRefined` class.
  - Make the generation of the LLM response also be a decorator

- Remove the `QueryResponseErrorDB` and also put error info `QueryResponseDB` with the correct migrations to avoid data loss. Update all code that uses these tables. Update data_api code with new logic.

## How has this been tested?

- pytests
- dev setup
- docker compose

## Future tasks

1. Update playground to always show sources
2. Update chat manager flows

## Checklist


amiraliemami:
Rebased with archiving changes from main. Not very familiar with the archiving stuff, all tests pass so I hope it's working correctly?

tonyzhao6:
the tests for archive do pass but they are unit tests at the end of the day.

for the Content Search and AI Response functionalities, [here is a video](https://www.loom.com/share/af58be40df0644f0936a252bdb350a39?sid=8d875776-fd77-4c7b-bf5d-48edefe4db61) showing how they work.

core_backend/app/question_answer/models.py
```diff
@@ -204,90 +205,35 @@ async def save_query_response_to_db(
...
+    if type(response) is QueryResponse:
+        user_query_responses_db = QueryResponseDB(
+            query_id=user_query_db.query_id,
+            search_results=response.model_dump()["search_results"],
+            llm_response=response.model_dump()["llm_response"],
+            response_datetime_utc=datetime.now(timezone.utc),
+            debug_info=response.model_dump()["debug_info"],
+            is_error=False,
+        )
+    elif type(response) is QueryResponseError:
```
sidravi1, line 217: So we are keeping two schemas but just one db model. Would it make sense to consolidate schemas as well?

core_backend/app/question_answer/routers.py
```diff
@@ -16,23 +16,23 @@
...
 )
-from ..llm_call.process_output import check_align_score__after
+from ..llm_call.process_output import (
+    check_align_score__after,
+    generate_llm_response__after,
+)
 from ..users.models import UserDB
 from ..utils import create_langfuse_metadata, setup_logger
-from .config import N_TOP_CONTENT_FOR_RAG, N_TOP_CONTENT_FOR_SEARCH
+from .config import N_TOP_CONTENT
```
sidravi1, line 31: These may end up being different if we add a cross-encoder but fine for now

sidravi1 (commented):
Overall looks good! Will do a more detailed review tomorrow and then approve.

Thanks @amiraliemami

sidravi1 (approved):
Lgtm! Thanks @amiraliemami

.secrets.baseline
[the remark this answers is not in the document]
amiraliemami: Yeah, I think this is picking up secrets in the PyDantic field examples, like feedback-secret-key

core_backend/app/llm_call/process_output.py
```diff
@@ -36,14 +39,98 @@ class AlignScoreData(TypedDict):
     claim: str
 
 
+def generate_llm_response__after(func: Callable) -> Callable:
+    """
+    Decorator to generate the LLM response.
+
+    Only runs if the generate_llm_response flag is set to True.
```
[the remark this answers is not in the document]
amiraliemami: Thanks, just made sense ;)

core_backend/app/question_answer/models.py
amiraliemami: Yeah because QueryResponseError is for when we return an error 400 (guardrail failure) and has extra error info, and QueryResponse is for a successful call. We also use isinstance with QueryResponseError to skip guardrails if previous ones have failed.

PR #338, opened 2024-07-31
Track content sent for a query

sidravi1:
Reviewer: @amiraliemami 
Estimate: 10mins

---

## Ticket

Fixes: N/A - to enable chart on dashboard pg2

### Goal

Track what content was shared for a query and when

### Changes

New table

## How has this been tested?

Tests updated and run successfully

## Checklist

Fill with `x` for completed. 


core_backend/app/question_answer/routers.py
```diff
@@ -104,18 +104,23 @@ async def search(
         exclude_archived=exclude_archived,
     )
 
-    if type(response) is QueryResponse:
-        await save_query_response_to_db(user_query_db, response, asession)
```
sidravi1, line 108: We were running the same lines for both types so moved it out. This does mean that it will error if the type is not `QueryResponse` or `QueryResponseError` even before we get to the ValueError line

core_backend/migrations/versions/2024_07_31_8946f1949a33_add_contents_for_query_table.py
```diff
@@ -0,0 +1,57 @@
...
+            ["query_id"],
+            ["query.query_id"],
+        ),
+        sa.ForeignKeyConstraint(
+            ["user_id"],
+            ["user.user_id"],
+        ),
+        sa.PrimaryKeyConstraint("content_for_query_id"),
+    )
+    op.create_index(
```
sidravi1, line 44: Since this is what we'll use to query

core_backend/app/question_answer/routers.py
sidravi1: Though in theory that should never happen

core_backend/app/contents/models.py
```diff
@@ -229,14 +273,19 @@ async def update_content_in_db(
 
 
 async def increment_query_count(
```
amiraliemami, line 275: We should do the updating the content history table in a separate function (esp if we move it to the AQ module) or change this function name and docstring to reflect its actions.

core_backend/app/question_answer/routers.py
amiraliemami: yeah fair enough... let's move the check up then (like if not either of the two types return ValueError, else move on)? Or drop it altogether?

core_backend/app/contents/models.py
```diff
@@ -116,6 +113,53 @@ def __repr__(self) -> str:
         )
 
 
+class ContentForQueryDB(Base):
+    """
+    ORM for storing what content was returned for a given query.
+    Allows us to track how many times a given content was returned in a time period.
+    """
+
+    __tablename__ = "content_for_query"
```
amiraliemami, line 122: btw, we decided to use dashes for table names instead of underscores

core_backend/app/contents/models.py
```diff
@@ -116,6 +113,53 @@ def __repr__(self) -> str:
         )
 
 
+class ContentForQueryDB(Base):
```
amiraliemami, line 116: Not sure about the location of this + its class name + table name, doesn't seem clear to me... I'm wondering if this should actually be defined beside QueryResponseDB inside the question_answer module and be called QueryResponseContentDB. This would logically make sense (these records are added only after a search is done) and the name would be clear too.


If it HAS to stay here, about  ContentReturnedDB or ContentSearchHistoryDB or ContentSearchResultDB or smt like that? With the corresponding SQL table name. I'd vote for moving it though.

core_backend/app/contents/models.py
```diff
@@ -116,6 +113,53 @@ def __repr__(self) -> str:
...
+    )
+    user_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("user.user_id"), nullable=False
+    )
+    content_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("content.content_id"), nullable=False
+    )
+    query_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("query.query_id"), nullable=False
+    )
```
amiraliemami, line 135: let's swap these columns?

```suggestion
    query_id: Mapped[int] = mapped_column(
        Integer, ForeignKey("query.query_id"), nullable=False
    )
    content_id: Mapped[int] = mapped_column(
        Integer, ForeignKey("content.content_id"), nullable=False
    )
```

amiraliemami (changes requested):
Thanks for this, left a few comments/suggestions!

core_backend/app/contents/models.py
sidravi1: Leaving this as in as discussed in the tech session

core_backend/app/contents/models.py
sidravi1: Make sense. I'll move it

core_backend/app/question_answer/routers.py
sidravi1: Ok will do

core_backend/app/question_answer/models.py
```diff
@@ -234,6 +236,77 @@ async def save_query_response_to_db(
...
+        DateTime(timezone=True), nullable=False
+    )
+
+    __table_args__ = (
+        Index("idx_user_id_created_datetime", "user_id", "created_datetime_utc"),
+    )
+
+    def __repr__(self) -> str:
+        """
+        Construct the string representation of the `ContentForQueryDB` object.
```
tonyzhao6, line 269: it's for the `QueryResponseContentDB` object.

core_backend/app/question_answer/models.py
```diff
@@ -234,6 +236,77 @@ async def save_query_response_to_db(
...
+
+        Returns
+        -------
+        str
+            A string representation of the `ContentForQueryDB` object.
```
tonyzhao6, line 274: it's for the `QueryResponseContentDB` object.

core_backend/app/question_answer/models.py
```diff
@@ -8,13 +8,14 @@
 """
 
 from datetime import datetime, timezone
-from typing import List
+from typing import Dict, List
```
tonyzhao6, line 11: can we start using native types instead of importing from `typing`?

tonyzhao6 (approved):
Minor cosmetic change if you want to incorporate, otherwise lgtm.

Local tests pass but understand there is an ongoing issue with tests in GA.

core_backend/app/question_answer/models.py
sidravi1: TIL that importing these from typing is [deprecated](https://peps.python.org/pep-0585/). Sounds good!

PR #371, opened 2024-08-09
Added guards to endpoint

sidravi1:
Reviewer: @suzinyou 
Estimate: 5 mins

---

## Ticket

Fixes: N/A

### Goal

Adding back guards to check env vars are set before running a target

## Checklist

Fill with `x` for completed. 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ make teardown-redis:
 
 
 # Dev LiteLLM Proxy server
-setup-llm-proxy: guard-OPENAI_API_KEY guard-CUSTOM_EMBEDDINGS_API_KEY guard-CUSTOM_EMBEDDINGS_ENDPOINT guard-VERTEXAI_PROJECT guard-VERTEXAI_LOCATION
+setup-llm-proxy:
 	-@docker stop litellm-proxy
 	-@docker rm litellm-proxy
 	@docker system prune -f
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ build-embeddings-arm:
 	@cd ..
 	@rm -rf text-embeddings-inference
 
-setup-embeddings-arm: guard-HUGGINGFACE_MODEL guard-CUSTOM_EMBEDDINGS_API_KEY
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -69,7 +71,7 @@ make teardown-redis:
 
 
 # Dev LiteLLM Proxy server
-setup-llm-proxy:
+setup-llm-proxy: guard-OPENAI_API_KEY guard-CUSTOM_EMBEDDINGS_API_KEY guard-CUSTOM_EMBEDDINGS_ENDPOINT guard-VERTEXAI_PROJECT guard-VERTEXAI_LOCATION
```
suzinyou, line 74: Can we remove these? we pass .env file instead now!

suzinyou (commented):
Thanks for this Sid!

With the new change, we are trying to consolidate all configurations in .*.env files, and indicating what's required vs. not by having env variables commented out for optional ones.

So we should actually pass `--env-file .litellm_proxy.env` to the `docker run`'s in embeddings targets..

But I see that in that case, we can't catch before running containers when a required env var hasn't been defined in that .env file..

Makefile
suzinyou, line 74:
```suggestion
setup-llm-proxy:
```

suzinyou (approved):
Could you make sure there are no more conflicts and merge? thank you!

Makefile
```diff
@@ -103,7 +105,8 @@ build-embeddings-arm:
 		-t text-embeddings-inference-arm
 	@cd ..
 	@rm -rf text-embeddings-inference
-setup-embeddings-arm:
+
+setup-embeddings-arm: guard-HUGGINGFACE_MODEL guard-CUSTOM_EMBEDDINGS_API_KEY
```
suzinyou, line 109: Ah, the alternative is to export env vars from .env file within the make target!

But user needs to remember to set them.

Because I want this for momconnect I will just push it directly to main :'( Sorry for the mess

```
Ah! The alternative is this.
	@set -a && \
        source "$(CURDIR)/deployment/docker-compose/.core_backend.env" && \
        set +a && \
	docker run \
            ....
```

suzinyou: Oh nooo it doesn't actually work. I've tested. Nvm.

PR #395, opened 2024-08-14
[128] - Refactoring of the E2E speech endpoint

MustafaAkolawala:
Reviewer: @amiraliemami 
Estimate: 20 min

---

## Ticket

Fixes: [JIRA_TICKET_LINK](https://idinsight.atlassian.net/browse/AAQ-767?atlOrigin=eyJpIjoiYWNmNTRhYzYwYzc5NDQ5OGI4ZGMxOTg5NDVlNGQxZmIiLCJwIjoiaiJ9)

## Description
This PR implements a new end-to-end speech workflow design via the `voice-search` endpoint to make the . This change isolates speech functionality, making it more intuitive and easier to implement

### Goal
To create a dedicated, isolated endpoint for the end-to-end speech workflow, featuring an optional `generate_tts` flag. This flag enables the generation of voice notes from the LLM response, streamlining speech functionality implementation and improving overall intuitiveness.

### Changes

- Introduced a new Pydantic model `AudioResponse`, inheriting from `QueryResponse`.
- Implemented the `voice-search` endpoint, capable of processing voice files and optionally returning voice responses.
- Created a `generate_tts__after` decorator, which requires an `llm-response` in the response object to execute.

## How has this been tested?
dev environment
docker-compose
unit Tests

### How to test this?
1. Configure the `Speech_Api` environment variables in `template.core_backend.env`.
2. Launch the Docker containers using:  `docker compose -f docker-compose.yml -f docker-compose.dev.yml -f docker-compose.speech.yml -p  aaq-stack watch`
3. Send a POST request to the `voice-search` endpoint.

Once `VoiceApi/GCP` is merged, will merge this to main

## Checklist

Fill with `x` for completed. 


deployment/docker-compose/docker-compose.speech.yml
```diff
@@ -7,6 +7,7 @@ services:
       - ./.gcp_credentials.json:/app/credentials.json
     env_file:
       - .litellm_proxy.env
+
```
amiraliemami, line 10: btw, I don't think you need the `ports: - "8001:8001"` bit since we're accessing it using `speech_service:8001` as the URL - no need to expose the port through the host machine (which is what this does) since it's accessed from within the docker network

core_backend/Makefile
```diff
@@ -9,7 +9,7 @@ tests: setup-test-containers run-tests teardown-test-containers
 # NB: `pytest-alembic` requires the DB engine to point an empty database. Thus, alembic
 # tests should be run first.
 run-tests:
-	@set -a && source ./tests/api/test.env && set +a && \
+	@set -a && . ./tests/api/test.env && set +a && \
```
amiraliemami, line 12: why did this change?

core_backend/app/voice_api/voice_components.py
```diff
@@ -12,13 +12,17 @@
...
     destination_blob_name: str = "response.mp3",
 ) -> str:
     """
     Converts the provided text to speech and saves it as an mp3 file on
     Google cloud storage
     """
+    if language is None:
+        error_msg = "Language must be provided to generate speech."
+        logger.error(error_msg)
+        raise ValueError(error_msg)
```
amiraliemami, line 25: I'd say keep the language parameter not-nullable in this function and move this check to the `_generate_tts_response` that calls it. This function should only be called if all of its parameters are available

core_backend/app/question_answer/schemas.py
```diff
@@ -90,6 +90,16 @@ class QueryResponse(BaseModel):
     model_config = ConfigDict(from_attributes=True)
 
 
+class AudioResponse(QueryResponse):
+    """
+    Pydantic model for response to a Voice Query with audio response and Text response
+    """
+
+    tts_file: str | None = Field(None, examples=["response.mp3"])
```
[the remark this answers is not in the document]
amiraliemami: also maybe we make the example look like a fake Gcloud bucket URL since that's what it will be?

core_backend/app/question_answer/routers.py
```diff
@@ -79,28 +87,38 @@
         },
     },
 )
-async def stt_llm_response(
-    generate_tts: bool = Form(...),
+async def voice_search(
+    generate_tts: bool = Form(False),
```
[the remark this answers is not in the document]
amiraliemami: Actually, my radical view is that we should remove this option. We should always return a voice file if the user's input was voice. Otherwise they can use another transcription API (or we can expose ours) and then pass the text to our normal /search endpoint. 

I think voice-to-voice would be the expected behaviour from this endpoint and it's what we need for the demo. 

Perhaps remove this and also response model will always be AudioResponse. If any error occurs, we should return an error 500 with what we managed to get so far

core_backend/app/question_answer/routers.py
```diff
@@ -155,9 +156,11 @@ async def stt_llm_response(
         os.remove(file_path)
         file_stream.close()
 
-    if type(response) is QueryResponse:
+    if isinstance(response, AudioResponse):
         return response
-    elif type(response) is QueryResponseError:
+    elif isinstance(response, QueryResponse):
+        return response
+    elif isinstance(response, QueryResponseError):
```
amiraliemami, line 163: are you sure isinstance() works here? It didn't work for /search which is why I used the if type() is XX format.

core_backend/app/question_answer/routers.py
```diff
@@ -253,6 +257,7 @@ async def search(
 @classify_safety__before
 @translate_question__before
 @paraphrase_question__before
+@generate_tts__after
 @generate_llm_response__after
 @check_align_score__after
```
amiraliemami, line 262: wait what was the conclusion here? Shouldn't it be:

```
@generate_tts__after
@check_align_score__after
@generate_llm_response__after
```

amiraliemami (commented):
Awesome work man, you understand the codebase inside out! Left some comments for now just looking at the code. Will test the functionality in a bit

core_backend/app/question_answer/routers.py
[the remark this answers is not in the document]
MustafaAkolawala: makes sense! i will keep it explicitly voice to voice then

core_backend/app/llm_call/process_output.py
```diff
@@ -303,3 +277,91 @@ async def _get_llm_align_score(
...
+
+        response = await func(query_refined, response, *args, **kwargs)
+
+        if not query_refined.generate_tts:
+            return response
+
+        if not isinstance(response, AudioResponse):
+            logger.warning(
+                f"Converting response of type {type(response)} to AudioResponse."
+            )
```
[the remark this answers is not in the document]
MustafaAkolawala: oh, i had kept that as logger.info i donno why is it logger.warning again, but yea it was just some extra info wont need it anymore cause it is explicitly speech E2E

core_backend/app/question_answer/schemas.py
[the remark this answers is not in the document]
MustafaAkolawala: yes makes sense will rename it

core_backend/app/voice_api/voice_components.py
```diff
@@ -12,13 +12,17 @@
 
 async def generate_speech(
     text: str,
-    language: IdentifiedLanguage,
+    language: IdentifiedLanguage | None,
     destination_blob_name: str = "response.mp3",
 ) -> str:
     """
     Converts the provided text to speech and saves it as an mp3 file on
```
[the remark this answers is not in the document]
MustafaAkolawala: i have that docstring properly written in the next external google cloud TTS and STT PR so that would work i hope!

core_backend/Makefile
MustafaAkolawala: OH so sorry, i by mistakenly comitted that change! thanks for pointing it out, so as am using WSL, as the main OS is still windows, the `source` command does not work for me so to run the makefiles i need to replace the `source` command

core_backend/app/question_answer/routers.py
MustafaAkolawala: oh it didn't? atleast while testing it was working, i was using the type( ) only but during the hurdle sid pointed it out that instead of type we should use isinstance cause that is better

core_backend/app/question_answer/routers.py
MustafaAkolawala: Sid did tell me that the` after` decorators work in the reverse order, but as align score was not being accessed at all due to some issue which ig carlos figured it out today, i did not change it for now but yes the `align score` decorator should be placed above the` llm_response` one

amiraliemami (approved):
Thanks! Great work and collab on this

PR #399, opened 2024-08-19
[AAQ-765] Retry LLM  generation when AlignScore fails

lickem22:
Reviewer: @amiraliemami 
Estimate: 40mins

---

## Ticket

Fixes:[AAQ-675](https://idinsight.atlassian.net/browse/AAQ-765)

### Goal
The goal of this PR is to allow retrying LLM response again when AlignScore fails because of a low score N times (default is 0). 
### Changes
The following changes have been made: 
- Add backoff library dependency
- Updated endpoint to retry when response is of type `QueryResponseError` and the error is low alignment score. Also allowed to add the previous failure raison in `response.debug_info["past_failure"]`
## How has this been tested?
Testing this is tricky because for this change to be observed we need LLM response to work but AlignScore to fail, and finding these cases are not straighforward. 
Was tested two ways:
First way is :
- Set `ALIGN_SCORE_THRESHOLD` to an unrealistic score (example 1.5). That way AlignScore fails
- Run `/search ` with `generate_lllm_response` set to `true`.  with a content and a question relevant for the content. An example is content: `Here we are going to talk about pineapples because of their pine shapes and the applee like tast`
 and question: `Are apple related to pineapples`,
- Make sure LLM response is ran twice (by checking logs) if ALIGN_SCORE_N_RETRIES is set to the default value (1)
- Make sure debug_info["past_failure"] is in the returned response.
Second way was by still setting  `ALIGN_SCORE_THRESHOLD` to a value >1 but then adding a logic in the code to make sure the value of  `ALIGN_SCORE_THRESHOLD` is reduced to a reasonable value everytime the LLM response is regenerated to make sure that after the second retry the AlignScore check passes. So, in second run, `ALIGN_SCORE_THRESHOLD` should be less than 0.8. This approach is not straightforward. I am open to more efficient ways of testing this feature.

## Checklist

Fill with `x` for completed. 


core_backend/app/question_answer/routers.py
```diff
@@ -235,6 +237,17 @@ async def search(
...
+        failure_reason = response.debug_info["factual_consistency"]
+        response = await retry_search(
+            query_refined=user_query_refined_template,
+            response=response_template,
+            user_id=user_db.user_id,
+            n_similar=int(N_TOP_CONTENT),
+            asession=asession,
+            exclude_archived=True,
+        )
+        response.debug_info["past_failure"] = failure_reason
```
lickem22, line 250: Added that in case it works after the second try to  understand why it failed first.

core_backend/app/question_answer/routers.py
```diff
@@ -307,6 +320,39 @@ async def search_base(
...
+    by low alignment score.
+    """
+    return (
+        isinstance(response, QueryResponseError)
+        and response.error_type == ErrorType.ALIGNMENT_TOO_LOW
+    )
+
+
+@backoff.on_predicate(
+    backoff.expo,
```
lickem22, line 335: What `backoff.expo` does is basically waiting a little more everytime the function is reran in an exponential way just to handle the load better.

core_backend/app/question_answer/routers.py
```diff
@@ -307,6 +320,39 @@ async def search_base(
...
+    max_tries=int(ALIGN_SCORE_N_RETRIES),
+    predicate=is_unable_to_generate_response,
+)
+async def retry_search(
```
lickem22, line 339: Decided to try the whole search again because the failure can come from any guardrail, so instead of rerunning only llm generation it's best to run everything again.

suzinyou: I see how tricky it can be to just retry response generation + align score in the current setup.. 

How about we break up the `search_base` function into 
1. `search_pipeline`, with all the input check (__before) guardrails
2. `generation_pipeline` with generation + all the output check (__after) guardrails?

Like the following:
```python
@identify_language__before
@classify_safety__before
@translate_question__before
@paraphrase_question__before
async def search_pipeline(  
    query_refined: QueryRefined,
    response: QueryResponse,
    user_id: int,
    n_similar: int,
    asession: AsyncSession,
    exclude_archived: bool = True,
) -> QueryResponse | QueryResponseError:
    # always do the embeddings search even if some guardrails have failed
    metadata = create_langfuse_metadata(query_id=response.query_id, user_id=user_id)
    search_results = await get_similar_content_async(
        user_id=user_id,
        # use latest version of the text
        question=query_refined.query_text,
        n_similar=n_similar,
        asession=asession,
        metadata=metadata,
        exclude_archived=exclude_archived,
    )
    response.search_results = search_results

    return response


@generate_tts__after
@backoff.on_predicate(
    backoff.expo,
    max_tries=int(ALIGN_SCORE_N_RETRIES),
    predicate=is_unable_to_generate_response,
)  # <--- so that we don't run generate_tts unnecessarily.
@check_align_score__after
@generate_llm_response__after
async def generation_pipeline(
    query_refined: QueryRefined,
    response: QueryResponse,
    user_id: int,
    n_similar: int,
    asession: AsyncSession,
    exclude_archived: bool = True,
) -> QueryResponse | QueryResponseError:
    # Maybe put the `generate_llm_response__after` code here?
    return response
```

We'll probably have to change how we handle `generate_llm_response` inside `generate_llmresponse__after`.. Or as I mention above, just move the `generate_llm_response__after` code into `generation_pipeline.

core_backend/requirements.txt
```diff
@@ -19,4 +19,5 @@ types-openpyxl==3.1.4.20240621
 redis==5.0.8
 python-dateutil==2.8.2
 gTTS==2.5.1
-google-cloud-storage==2.18.2
+backoff==2.2.1
+google-cloud-storage==2.18.2
```
suzinyou, line 23: There should be a new line at the end of every file. This should be caught be pre-commit hooks!

```suggestion
google-cloud-storage==2.18.2

```

core_backend/app/question_answer/routers.py
lickem22: Thanks @suzinyou ,
1) What I meant is that, redoing each step might actually improve the alignScore for example:
- A new paraphased text can improve the LLM response and thus improve alignScore
- A new translation can improve the LLM response and so on 
That is the reason one I think it is beneficial to run the whole pipeline again. 
2) I do agree that rerunning search is wasteful but the whole reason why we are using decorators is to unify search and llm_response. Wouldn't what you suggest bring us backwards? I think this is worth discussing in a tech session before I actually implement the changes.

suzinyou: 1. Taking this to slack
2. I think it's actually quite opaque how we handle generate_llm_response inside a decorator, so it would be good to bring it out to the router logic. It is easier to understand, and we still keep the checks and transformations (guardrails) as decorators.

core_backend/app/llm_call/llm_prompts.py
```diff
@@ -177,6 +177,11 @@ def get_prompt(cls) -> str:
 You are a helpful question-answering AI. You understand user question and answer their \
 question using the REFERENCE TEXT below.
 """
+RETRY_PROMPT_SUFFIX = """
```
lickem22, line 180: Added a suffix to the prompt to incorporate failure reason

core_backend/app/llm_call/llm_prompts.py
```diff
@@ -177,6 +177,11 @@ def get_prompt(cls) -> str:
...
+If the response above is not aligned with the question, please rectify this by \
+considering the following reason(s) for misalignment: "{failure_reason}". 
+Make necessary adjustments to ensure the answer is aligned with the question.
+"""
```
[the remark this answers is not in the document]
suzinyou: Also, shouldn't the prompt define what we mean by alignment?

suzinyou (commented):
Thanks Carlos! I think we need to think and validate the prompt a bit. Should we discuss in the next tech session?

core_backend/app/question_answer/routers.py
lickem22: I guess it depends on how useful the approach is, since we haven't done any analysis to see how well it works.
But personnally I think since it doesn't add a dependency (backoff being used by litellm), and since the only code we would change if we retry just once is the decorator and the config variable, the cost is pretty low, so we can just keep it.

core_backend/app/llm_call/llm_rag.py
```diff
@@ -37,7 +37,14 @@ async def get_llm_rag_answer(
     """
 
     metadata = metadata or {}
-    prompt = RAG.prompt.format(context=context, original_language=original_language)
+    if "failure_reason" in metadata and metadata["failure_reason"]:
```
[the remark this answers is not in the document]
lickem22: My initial understanding was that we are using this to try the functionality. What if we keep it like this while testing, and if it turns out to be something we want to keep, the we will explicitly set it as a functionality by addind the `is_retry` parameter. What do you think?

core_backend/app/llm_call/llm_prompts.py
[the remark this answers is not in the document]
lickem22: That makes sense. To be honest,  I was just having a go at updating the prompt to take the output into consideration. I am not exactly an expert in prompt engineering. Should we discuss that in a tech session?

PR #403, opened 2024-08-21
[128] - Google cloud TTS and STT functionality

MustafaAkolawala:
Reviewer: @lickem22 
Estimate: 20 mins

---

## Ticket

Fixes: [JIRA_TICKET_LINK](https://idinsight.atlassian.net/browse/AAQ-768?atlOrigin=eyJpIjoiMzRkYjdkMDcwNDRkNGM0NmIxOWYzMTQ4MmViZDA0MTIiLCJwIjoiaiJ9)

## Description
This pull request  implements the Google Cloud `text-to-speech (TTS`) and `speech-to-text (STT)` external APIs in the existing AAQ  workflow, alongside the existing in-house models.

### Goal
The goal of this PR is to provide users with the option to use either the in-house models or the new Google Cloud `TTS` and `STT` models, depending on their specific use cases and requirements.

### Changes
- Added the ability to select between external APIs or in-house models based on the `CUSTOM_SPEECH_ENDPOINT` environment variable.
- introduced new external voice components in the `core_backend/app/question_answer/speech_components` directory.

### Future Tasks 
- Determine the best approach to detect the language for the external STT model.
- Explore the possibility of an automated way to map languages to their corresponding language codes.
 
## How has this been tested?
docker-compose
swagger UI
Unit Tests (pytests)

### How to test this?
1. Ensure that the `CUSTOM_SPEECH_ENDPOINT` environment variable is not set, to access the external APIs, and that your `GCP credentials `are properly configured.
2. Run the command `docker compose -f docker-compose.yml -f docker-compose.dev.yml -p aaq-stack watch` (the `speech_service` container is not needed for the external APIs).
3. Test the functionality by accessing the endpoint and viewing the generated file through the `generated URL`.

## Checklist

Fill with `x` for completed. 


core_backend/app/llm_call/process_output.py
```diff
@@ -341,7 +346,6 @@ async def _generate_tts_response(
         logger.warning("TTS generation skipped due to missing LLM response.")
         return response
```
lickem22, line 348: Must have lost some context, why are we not checking for the language?

MustafaAkolawala: here we are just checking if the LLM gave us a language or not as in the `IdentifiedLanguage` should never be none, then we map the appropriate codes inside the speech function itself

core_backend/app/question_answer/speech_components/utils.py
```diff
@@ -0,0 +1,68 @@
...
+def get_gtts_lang_code_and_model(
+    identified_language: IdentifiedLanguage,
+) -> tuple[str, str]:
+    """
+    Maps IdentifiedLanguage values to Google Cloud Text-to-Speech language codes
+    and voice model names.
+    """
+    # Add language codes and voice models according to
+    # https://cloud.google.com/text-to-speech/docs/voices
+    mapping = {
```
lickem22, line 20: You shouls probably put it outside the function and like this 
````
LANG_CODE_MAPPING = ....
````

File: core_backend/app/question_answer/speech_components/utils.py [hunk 3]
```diff
@@ -0,0 +1,68 @@
...
+
+    if file_extension in ["mp3", "m4a"]:
+        audio = AudioSegment.from_file(input_filename, format=file_extension)
+
+        wav_filename = os.path.splitext(input_filename)[0] + ".wav"
+        audio.export(wav_filename, format="wav")
+        logger.info(f"Conversion complete. Output file: {wav_filename}")
+    else:
+        wav_filename = input_filename
+        logger.info(f"{wav_filename} is already in WAV format.")
```
lickem22, line 49: You should probably handle the case where it is not wav. Or are we assuming that if it's neither in mp3 or m4a that means it is in wav by default?

MustafaAkolawala: yeah i was thinking that these 3 are the major types anyways right? are there any others we should consider?

lickem22: With the current code, an mp4 file would be considered as wav. We should handle the case where it is not part of any of the 3 format and  probably return an error.

core_backend/app/question_answer/routers.py
```diff
@@ -108,18 +111,31 @@ async def voice_search(
...
+            query_text=transcription_result["text"],
+            query_metadata={},
+        )
+    else:
+        external_transcription_result = await transcribe_audio(file_path)
+        user_query = QueryBase(
+            generate_llm_response=True,
+            query_text=external_transcription_result,
+            query_metadata={},
+        )
```
lickem22, line 137:
```suggestion
    if CUSTOM_SPEECH_ENDPOINT != "":
        transcription = await post_to_speech(file_path, CUSTOM_SPEECH_ENDPOINT)
        transcription_result = transcription["text"]
        
    else:
        transcription_result = await transcribe_audio(file_path)
    user_query = QueryBase(
            generate_llm_response=True,
            query_text=transcription_result
            query_metadata={},
        )
```

lickem22: So that you dont have to repeat `user_query`

deployment/docker-compose/docker-compose.yml
```diff
@@ -7,14 +7,19 @@ services:
     command: >
       /bin/sh startup.sh
     restart: always
+    volumes:
+      - temp:/usr/src/aaq_backend/temp
+      - ./.gcp_credentials.json:/app/credentials.json
     env_file:
       - .base.env
       - .core_backend.env
+      - .litellm_proxy.env
```
lickem22, line 16: Why adding `.litellm_proxy.env`

deployment/docker-compose/docker-compose.yml
```diff
@@ -7,14 +7,19 @@ services:
     command: >
       /bin/sh startup.sh
     restart: always
+    volumes:
+      - temp:/usr/src/aaq_backend/temp
```
lickem22, line 11: What is that for?

core_backend/app/config.py
```diff
@@ -69,9 +69,7 @@
 BACKEND_ROOT_PATH = os.environ.get("BACKEND_ROOT_PATH", "")
 
 # Speech API
-SPEECH_ENDPOINT = os.environ.get(
-    "SPEECH_ENDPOINT", "http://speech_service:8001/transcribe"
-)
+CUSTOM_SPEECH_ENDPOINT = os.environ.get("CUSTOM_SPEECH_ENDPOINT", "")
```
lickem22, line 72: You can just use None,

core_backend/app/question_answer/speech_components/utils.py
MustafaAkolawala: makes sense, i will do that

deployment/docker-compose/docker-compose.yml
MustafaAkolawala: so that we can use the `Google cloud storage buckets` and also use these `external apis` without explicitly running the speech container

deployment/docker-compose/docker-compose.yml
MustafaAkolawala: that was for creating a temporary local copy of the `audio file` on our system for transcription purposes cause whisper needs the file locally somewhere for transcription, we later on delete it

lickem22 (approved):
Thanks Mustafa, 
Added a few comments but looks good overall!

deployment/docker-compose/docker-compose.yml
lickem22: I see. I do believe that if we are using some variables in more than one container we should put it in .base.env. But that's a discussion for another day.

PR #414, opened 2024-08-27
Paraphrase prompt fix & cross encoder

sidravi1:
Reviewer: @suzinyou 
Estimate: 15mins

---

## Ticket

Fixes: [This issue](https://idinsight.slack.com/archives/C05RSDLN1GA/p1724774049745359)

### Goal
Improve performance

### Changes
- Improves paraphrasing prompt
- Adds a cross-encoder that can be disabled if needed

## How has this been tested?
Test pass

## Checklist

Fill with `x` for completed. 

### DIFF
--- a/core_backend/app/__init__.py
+++ b/core_backend/app/__init__.py
@@ -7,6 +7,7 @@ from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 from prometheus_client import CollectorRegistry, make_asgi_app, multiprocess
 from redis import asyncio as aioredis
+from sentence_transformers import CrossEncoder
 
 from . import (
     admin,
@@ -20,7 +21,7 @@ from . import (
     urgency_rules,
     user_tools,
 )
-from .config import DOMAIN, LANGFUSE, REDIS_HOST
+from .config import CROSS_ENCODER_MODEL, DOMAIN, LANGFUSE, REDIS_HOST, USE_CROSS_ENCODER
 from .prometheus_middleware import PrometheusMiddleware
 from .utils import setup_logger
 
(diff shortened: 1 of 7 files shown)


suzinyou (changes requested):
Hey sid, thanks for this quick attempt! 

What value did you use for the `SERVICE_IDENTITY` environment variable? 
We had used it for on/off topic check, but it was very brittle so we removed it.
Also, it's a global variable and can't be configured per user, so its value needs to be use-case agnostic. So it's probably better to just bake it into the paraphrase prompt.

Also, we'll need to update the CI/CD scripts to set any new env variables..

It would also be nice to check if `USE_CROSS_ENCODER == "True"` then check that `N_TOP_CONTENT_TO_CROSSENCODER >= N_TOP_CONTENT`

(For future, I wonder if we can move crossencoder to the embedding server or litellm..)

core_backend/app/__init__.py
```diff
@@ -5,6 +5,7 @@
 from fastapi.middleware.cors import CORSMiddleware
 from prometheus_client import CollectorRegistry, make_asgi_app, multiprocess
 from redis import asyncio as aioredis
+from sentence_transformers import CrossEncoder
```
suzinyou, line 8: We can address this at a later point but getting this error 
>/usr/local/lib/python3.10/site-packages/transformers/tokenization_utils_base.py:1601: FutureWarning: `clean_up_tokenization_spaces` was not set. It will be set to `True` by default. This behavior will be depracted in transformers v4.45, and will be then set to `False` by default. For more details check this issue: https://github.com/huggingface/transformers/issues/31884

suzinyou (approved):
Works well, thanks!!

core_backend/app/__init__.py
sidravi1: I tried to see how to fix this but it wasn't obvious. I'll do it in a different PR.

PR #443, opened 2024-10-02
[AAQ-835] Enable LLM-less dashboards

markbotterill:
Reviewer: @lickem22 
Estimate: 20 mins

---

## Ticket AAQ-835

Fixes: [JIRA_TICKET_LINK](https://idinsight.atlassian.net/browse/AAQ-835?atlOrigin=eyJpIjoiNzRkNWJhNjNlMDAzNDU3OWJhYTRlZmI3OTVmMWQzNjYiLCJwIjoiaiJ9)

## Description

Introduces new .env variable `DISABLE_DASHBOARD_LLM` to toggle P3 behavior. Replace `AI_SUMMARY` with this new variable to achieve similar effect. 

### Goal

First step to making an easy-switch LLM-less AAQ

### Changes

Changes in backend topic modeling pipeline for topic summaries. 

Small frontend bug fixed ("Run discovery" button was overflowing across lines") + enabled line formatting for one frontend box

### Future Tasks (optional)

Curate necessary changes required for `DISABLE_DASHBOARD_LLM` -> `DISABLE_LLM` to be app-wide. Execute changes. 

## How has this been tested?

Runs locally + passes GHA tests

## Checklist

Fill with `x` for completed. 

- [x ] My code follows the style guidelines of this project
- [ x] I have reviewed my own code to ensure good quality
- [ x] I have tested the functionality of my code to ensure it works as intended
- [x ] I have resolved merge conflicts

### DIFF
--- a/core_backend/app/dashboard/topic_modeling.py
+++ b/core_backend/app/dashboard/topic_modeling.py
@@ -3,6 +3,7 @@ This module contains functions for the topic modeling pipeline.
 """
 
 import asyncio
+import os
 from datetime import datetime, timezone
 from typing import Any, Coroutine, Dict, List, Tuple, cast
 
@@ -19,6 +20,11 @@ from .config import TOPIC_MODELING_CONTEXT
 from .schemas import BokehContentItem, Topic, TopicsData, UserQuery
 
 logger = setup_logger()
+
+# Check if LLM functionalities are disabled for dashboard
+DISABLE_DASHBOARD_LLM = (
+    os.environ.get("DISABLE_DASHBOARD_LLM", "false").lower() == "true"
+)
 
 
 async def topic_model_queries(
@@ -77,8 +83,8 @@ async def topic_model_queries(
     # Add reduced embeddings (for visualization)
     add_reduced_embeddings(results_df, topic_model)
 
-    # Generate topic labels using LLM
-    topic_labels = await generate_topic_labels_async(results_df, user_id)
+    # Generate topic labels using LLM or alternative method
+    topic_labels = await generate_topic_labels_async(results_df, user_id, topic_model)
 
     # Add topic titles to the DataFrame
     results_df["topic_title"] = results_df.apply(
@@ -161,9 +167,9 @@ def add_reduced_embeddings(results_df: pd.DataFrame, topic_model: BERTopic) -> N
 
 
 async def generate_topic_labels_async(
-    results_df: pd.DataFrame, user_id: int
+    results_df: pd.DataFrame, user_id: int, topic_model: BERTopic
 ) -> Dict[int, Dict[str, str]]:
-    """Generate topic labels asynchronously using an LLM."""
+    """Generate topic labels asynchronously using an LLM or alternative method."""
     tasks: List[Coroutine[Any, Any, Dict[str, str]]] = []
     topic_ids: List[int] = []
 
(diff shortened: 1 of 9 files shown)


core_backend/app/dashboard/topic_modeling.py
```diff
@@ -20,6 +21,11 @@
 
 logger = setup_logger()
```
lickem22, line 23: Why do you have this again since it's in the `config.py` file?

core_backend/app/llm_call/dashboard.py
```diff
@@ -19,6 +22,11 @@ async def generate_ai_summary(
     """
     Generates AI summary for the dashboard.
     """
+    if DISABLE_DASHBOARD_LLM:
+        logger.info("LLM functionality is disabled. Returning default message.")
+        return """LLM functionality disabled on the backend. Please check your
```
lickem22, line 27: I think logging is fine but I am not sure we should return a string when it's disabled. That would confuse the users since they can't easily tell between actual AI summary and the error text. Let's just return `None`. What do you think?

core_backend/app/llm_call/dashboard.py
```diff
@@ -40,13 +48,37 @@ async def generate_topic_label(
...
+        topic_title = ", ".join(top_keywords[:3])  # Use top 3 keywords as title
+        # Use all keywords as summary
+        # Line formatting looks odd since 'pre-wrap' is enabled on the frontend
+        topic_summary = f"""{" ".join(top_keywords)}
+
+Hint: To enable full AI summaries please set the DASHBOARD_LLM environment variable to "True" in your configuration."""  # noqa: E501
+        logger.info(
+            f"Generated topic label for topic_id {topic_id} using basic keywords."
+        )
+        return {"topic_title": topic_title, "topic_summary": topic_summary}
```
lickem22, line 79: What do you think about using a Schema instead of a dict?

lickem22 (commented):
Hey @markbotterill
The dashboard seems to be broken after the changes. Can we debug together when you back?

core_backend/app/llm_call/dashboard.py
markbotterill: The useful thing is that this gets shown in the front-end. So if people are exploring their topics I think it makes more sense to tell them that it's disabled than for them to just see an empty box.

markbotterill (commented):
> Why do you have this again since it's in the `config.py` file?
>  

Which `config.py` are you referring to?

core_backend/app/llm_call/dashboard.py
markbotterill: My only counterargument would be that we do slot both of these into a schema later on [here](https://github.com/IDinsight/ask-a-question/blob/e03110d54ea7a9d90bae41208fff3be76e74920a/core_backend/app/dashboard/schemas.py#L180) so it might overkill to create a smaller schema for just this dict + then have Topics include that smaller schema within it.

core_backend/app/dashboard/topic_modeling.py
markbotterill: Which `config.py` do you mean? I thought I'd replicated the logger flow of other dashboard functionalities but may be wrong

core_backend/app/llm_call/dashboard.py
lickem22: But is an empty string expected when the LLM is on? 
If that's the case, writing that makes sense. If not we should just return None and on the frontend decide what to tell users. For example, for some use cases, the LLM being off is a mistake and for some use cases it is a feature. And the message should change depending on whether it is a mistake or a feature. 
So my suggestion would be :
- Return None
-  Add a message on frontend depending on whether it is an expected feature or not.

lickem22 (approved):

PR #489, opened 2025-02-24
[EXPERIMENT] Make sure ApexCharts explicitly imports + uses Inter

markbotterill:
Reviewer: N/A
Estimate: 5 mins

### DIFF
--- a/admin_app/src/app/dashboard/components/overview/HeatMap.tsx
+++ b/admin_app/src/app/dashboard/components/overview/HeatMap.tsx
@@ -1,13 +1,7 @@
 "use client";
 import dynamic from "next/dynamic";
 import { ApexOptions } from "apexcharts";
-import { Inter } from "next/font/google";
-
-const inter = Inter({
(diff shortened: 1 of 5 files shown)


admin_app/src/app/dashboard/components/overview/HeatMap.tsx
```diff
@@ -1,7 +1,13 @@
 "use client";
 import dynamic from "next/dynamic";
-
 import { ApexOptions } from "apexcharts";
+import { Inter } from "next/font/google";
```
sidravi1, line 4: We may want to do this in layout.tsx way up in admin/src/app so it's all inter

markbotterill: @sidravi1 I've refactored as suggested. Font now comes from `fonts.ts` and then gets accessed as needed throughout 🙏  Is anyone able to test that this works for them as well?

PR #493, opened 2025-03-03
Check for user default workspace when logging in with google

lickem22:
Reviewer: @tonyzhao6 
Estimate: 10mins

---

## Ticket

Fixes: [AAQ-899](https://idinsight.atlassian.net/browse/AAQ-899)

### Goal
The goal of the PR is to fix the issue that prevent existing users from signin in using google.
### Changes
- Updated google signin logic to sign check for default workspace instead of specific workspace
## Checklist

Fill with `x` for completed. 

[AAQ-899]: https://idinsight.atlassian.net/browse/AAQ-899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core_backend/app/auth/routers.py
+++ b/core_backend/app/auth/routers.py
@@ -11,15 +11,15 @@ from ..config import DEFAULT_API_QUOTA, DEFAULT_CONTENT_QUOTA
 from ..database import get_sqlalchemy_async_engine
 from ..users.models import (
     UserNotFoundError,
-    check_if_user_exists_in_workspace,
     create_user_workspace_role,
     get_user_by_username,
+    get_user_default_workspace,
     get_user_role_in_workspace,
     save_user_to_db,
 )
 from ..users.schemas import UserCreate, UserRoles
 from ..utils import update_api_limits
-from ..workspaces.utils import create_workspace, get_workspace_by_workspace_name
+from ..workspaces.utils import create_workspace
 from .config import NEXT_PUBLIC_GOOGLE_LOGIN_CLIENT_ID
 from .dependencies import authenticate_credentials, create_access_token
 from .schemas import AuthenticatedUser, AuthenticationDetails, GoogleLoginData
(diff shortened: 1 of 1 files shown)


core_backend/app/auth/routers.py
```diff
@@ -216,47 +216,30 @@ async def authenticate_or_create_google_user(
...
+            assert user_role is not None and user_role in UserRoles, f"{user_role = }"
+            return AuthenticatedUser(
+                access_level="fullaccess",
+                user_role=user_role,
+                username=user_db.username,
+                workspace_name=workspace_db.workspace_name,
             )
 
-        # 7.
+        # 5.
```
tonyzhao6, line 235: Can we also update the docstring for this function so that it matches the new steps?

tonyzhao6:
Just to confirm my understanding of the issue:

1. Did this issue arise because there was a gap between commits when existing google users were created with the `_` character and then the new migration removed the `_` character?
2. If so, then this change will effectively check for an *existing* user's default workspace. If it was previously created with the `_` character, then we use that.

lickem22:
> Just to confirm my understanding of the issue:
> 
> 1. Did this issue arise because there was a gap between commits when existing google users were created with the `_` character and then the new migration removed the `_` character?
> 2. If so, then this change will effectively check for an _existing_ user's default workspace. If it was previously created with the `_` character, then we use that.

Yes exactly, and in general, if the user exists we should be looking for the default workspace not a specific one in case the user changes the workspace name.

tonyzhao6 (approved):
approved---please ensure that the function docstring is updated since we removed some steps.

core_backend/app/auth/routers.py
lickem22: Done!

PR #496, opened 2025-03-10
init version of DocMuncher(TM)

poornimaramesh:
Reviewer: @ziakhan04 
Estimate: ~1hr

---

## Ticket

Fixes: [JIRA_TICKET_LINK](https://idinsight.atlassian.net/browse/AAQ-904)

## Description
v1 of document ingestion

### Changes
1. Added document ingestion endpoint:
    - Currently supports only pdfs or zip files of pdfs (when creating zip files, make sure there are no MACOSX files in there -- it makes the background tasks stall -- haven't figured out why yet)
    - Uses Mistral AI for OCR from pdf to markdown
    - Uses Langchain to convert markdown into cards
    - Uses FastAPI background tasks to manage ingestion jobs
2. Get status based on task ID:
    - made this just in case (can deprecate / remove)
3. Get jobs based on status (i.e. in progress, success, failed, etc.)
    - Some of the jobs are created earlier, but finish later than tasks that are created later. So you can now simply do the loading modal by looking for "in_progress" tasks and if it returns a non-zero length list
    - It also returns the fraction of files of that type from the endpoint response, so you can make a progress bar
4. Get all jobs sorted by created datetime
    - can use this to display all jobs within the last 24 hours (redis times out and deletes jobs after 24 hours) 

### Future Tasks
- Add tests for these endpoints
- Fix the MACOSX issue
- Fix linting issues with Mistral :/

## How has this been tested?
1. Go to [Mistral AI](https://mistral.ai/fr), login with idinsight, copy the existing API or create a new one. Update `.core_backend.env` with it.
2. Spin up an instance, test uploading a pdf file / .zip file (shared on Slack) and check that it results in cards.

## Checklist

### DIFF
--- a/core_backend/app/docmuncher/dependencies.py
+++ b/core_backend/app/docmuncher/dependencies.py
@@ -51,6 +51,12 @@ async def create_tag_per_file(
     list
         The content tags.
     """
+    if not filename:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Filename is required",
+        )
+
     tag = TagCreate(tag_name=filename.upper())
     if not await is_tag_name_unique(
         asession=asession, tag_name=tag.tag_name, workspace_id=workspace_id
--- a/core_backend/app/docmuncher/dependencies.py
+++ b/core_backend/app/docmuncher/dependencies.py
@@ -3,7 +3,7 @@ import os
 from datetime import datetime, timezone
 from typing import List
 
-from fastapi import HTTPException, Request, UploadFile, status
+from fastapi import HTTPException, Request, status
 from langchain.text_splitter import MarkdownHeaderTextSplitter
 from langchain_core.documents import Document
 from mistralai import DocumentURLChunk, Mistral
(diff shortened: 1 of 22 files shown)


core_backend/app/docmuncher/dependencies.py
```diff
@@ -0,0 +1,315 @@
...
+from ..tags.schemas import TagCreate
+from ..utils import setup_logger
+from .schemas import DocIngestionStatusPdf, DocStatusEnum
+
+logger = setup_logger()
+MISTRAL_CLIENT = None
+JOB_KEY_PREFIX = "docmuncher_job_"
+
+
+def get_mistral_client() -> Mistral:
```
ziakhan04, line 24: It might be better to use dependency injection and provide the mistral client the route rather than having it as a globla thing. It is supposed to be easier to swap out keys and testing and stuff in the future.

Could think about doing it in the future though. Does not seem to be critical.

ziakhan04 (approved):
Thanks! Approving this PR. Works for me :)

Shared a minor comment wrt Mistral client dependency

ziakhan04 (commented):
Please also note that a Mypy test is failing

PR #531, opened 2025-04-16
Update docker-compose-dev.yml to avoid running npm i locally

ziakhan04:
Reviewer: @suzinyou
Estimate: 5 mins

## Test by
- Removing your node_modules
- Running make run-docker-compose-dev
- The should work even though you don't have any local node modules


suzinyou (approved):
Works!

PR #532, opened 2025-04-16
Add doc validation

ziakhan04:
Reviewer: @lickem22 
Estimate: 1 - 2 hours

---

## Ticket

Fixes: [TICKET](https://idinsight.atlassian.net/browse/AAQ-911)

## Description
I have added front-end and backend functionality to allow for card validation and archiving once cards have been uploaded.

### Goal
After cards have been uploaded, don't automatically make them available for search or on the UI. Instead, Push users to validate / archive cards before they appear in search or UI.

### Changes
- Added 2 columns to the DB content table - is_validated (used) and is_active (not used but can be useful in the future. It summarizes the status of each content and whether it will show up for search or not. We should use this but the refactor would have been way out of scope of ticket)
- Added a few backend routes - validate cards, get count of unvalidated cards, get next unvalidated card to allow for front-end stuff to work
- Made changes to a bunch of content page components
- Created new api calls in the front-end and also made tanstack wrappers for some existing api calls to make things simpler (they don't affect existing functionality. Only improves stuff in my ticket). Can refactor things in the future if needed.

### Future Tasks (optional)
- Write tests for the new api endpoints
- Use is_active in the backend to track content card status
- Modify the content card fetch function to use react query format to make things simpler and easier to manage
- Organize existing components in the codebase so that we don't have multiple components in the same module

## How has this been tested?
- Ran the code and tested it myself

## Checklist

Fill with `x` for completed. 

- x ] I have reviewed my own code to ensure good quality


core_backend/app/contents/models.py
```diff
@@ -112,6 +130,7 @@ async def save_content_to_db(
     asession: AsyncSession,
     content: ContentCreate,
     exclude_archived: bool = False,
+    exclude_unvalidated: bool = False,
```
lickem22, line 133: I am struggling to understand why adding this parameter. It's useful when saving but not when saving.

core_backend/app/contents/routers.py
```diff
@@ -306,12 +316,163 @@ async def retrieve_content(
     return contents
```
lickem22, line 318: Woudn't it be easy to handle that on the frontend rather than the backend? what do you think?

core_backend/app/contents/routers.py
```diff
@@ -306,12 +316,163 @@ async def retrieve_content(
...
+
+    await mark_content_as_validated(
+        asession=asession, content_id=content_id, workspace_id=workspace_id
+    )
+
+    return
+
+
+@router.get("/unvalidated-count")
+async def get_unvalidated_card_count(
```
lickem22, line 442: By doing what I suggested above, you wouldn't need this endpoint. You can just return the list of unvalidated cards, and when validated on frontend you can just bulk validate it at the end of the validation process instead of doing it one by one and calling the backend at each step.

admin_app/src/app/content/page.tsx
```diff
@@ -545,42 +559,69 @@ const CardsUtilityStrip: React.FC<CardsUtilityStripProps> = ({
...
-              startIcon={
-                data === true ? <CircularProgress size={12} color="inherit" /> : null
-              }
-            >
-              {data === true ? "Processing PDF" : "PDF Upload Status"}
-            </Button>
-            <IndexingStatusModal />
-          </>
-        )}
+        <>
```
lickem22, line 562: I like the fact that we are not using the same ContentViewModal object that we use to display contents would have made things complicated.

core_backend/app/contents/routers.py
ziakhan04: Replied generally in the other comment.

This is needed even before we fetch any unvalidated cards data. We could just load all unvalidated cards and then count them in the front end but that is not what the front-end needs. It just needs a count so the backend is only providing a count endpoint.

Just sharing what is needed.

core_backend/app/contents/models.py
ziakhan04: Yeah can remove. I guess not needed

ziakhan04 (commented):
> Hi @ziakhan04, thanks for this. I have two main comments:
> - Design comments you adressed I will  review them
> - There is a UI issue when here:
> Screenshot 2025-04-16 at 15 33 43
> when the select button are showing. Where do you think is the best place ti display it ? 

I can fix the view thanks for pointing it out.

I was thinking earlier that maybe when the select cards happens, other UI elements should disappear. I can do that. Hope that works?

PR #535, opened 2025-04-18
Limits for docmuncher

poornimaramesh:
Reviewer: @lickem22 
Estimate: ~10 minutes

---

## Ticket
Add upload limits for docmuncher

## Description
Set and enforce limits to restrict users from uploading too many documents, based on existing card limits
Here is roughly how it works:
1. Each time a document / zip file is uploaded, we count the number of pages in the document. We then make an _estimate_ of how many cards this will generate (I've set the default to 2 cards per page, but it is configurable via `.core_backend.env`).
2. We then update a variable in redis that keeps a running log of _estimated_ cards that will be generated by all the background processes
3. We then count the number of existing cards + estimated cards, and reject the current upload if this exceeds the set card limit

This ensures that (a) you cannot upload documents when you've exceeded the content limit, and (b) you can't simply run concurrent jobs to beat the content limit

### Changes
- Added content rate limiting for DocMuncher in the `docmuncher/upload` router.
- Added a env variable `PAGES_TO_CARDS_CONVERSION` in `template.core_backend.env` -- we need this to get estimated cards from each document
- Updated the frontend to display a full error message when an upload is rejected.

### Future Tasks (optional)
- Add unit tests
- Find a better way to get estimated cards from docs OR impose restrictions on number of pages also?

## How has this been tested?
1. In `.core_backend.env`, change the `ADMIN_CONTENT_QUOTA` to 10, and `PAGE_TO_CARDS_CONVERSION` to 2
2. Try uploading `test.pdf` and `test_2.pdf` from [this link](https://drive.google.com/drive/folders/16HCKxV4Hu4iFTscypnvwZNvejeZAD1VS) one immediately after the other (i.e. before the first job finishes). The second pdf should fail to upload because the content limit is exceeded.
3. Wait until the first job is finished and try uploading the second `test_2.pdf` file again. This upload should also fail
4. Delete any cards that were created. Now try uploading the zip file from the link above. The entire upload should fail.

## Checklist


core_backend/app/docmuncher/routers.py
```diff
@@ -16,17 +16,26 @@
...
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from ..auth.dependencies import get_current_user, get_current_workspace_name
-from ..config import REDIS_DOC_INGEST_EXPIRY_TIME as REDIS_EXPIRATION_SECONDS
+from ..config import (
+    CHECK_CONTENT_LIMIT,
+    PAGES_TO_CARDS_CONVERSION,
+)
+from ..config import (
+    REDIS_DOC_INGEST_EXPIRY_TIME as REDIS_EXPIRATION_SECONDS,
```
lickem22, line 28: Why the alias ?

core_backend/app/docmuncher/routers.py
```diff
@@ -135,7 +151,31 @@ async def upload_document(
...
 
-    # 2.
+    # 3.
+    # Get temporary log of expected contents to be created by running jobs
+    temp_docmuncher_contents = await redis.get("temp_docmuncher_contents")
+    if not temp_docmuncher_contents:
+        temp_docmuncher_contents = 0
+    else:
+        temp_docmuncher_contents = int(temp_docmuncher_contents.decode("utf-8"))
+    num_expected_contents = (
```
lickem22, line 161: There is an issue with that approach:
- As long as a document passes the import validation, the count is increased. And when it fails it's not updated an issue arises. I tried adding a doc after two failed because the tag existed And I got this error:
`An error occurred: Document ingestion could exceed content quota: Adding 12 new contents to the database would exceed the allowed limit of 9 contents.`.

core_backend/app/docmuncher/dependencies.py
```diff
@@ -455,6 +456,17 @@ async def process_pdf_file(
...
     finally:
         await redis.set(task_id, job_status_pydantic.model_dump_json())
+
+        temp_docmuncher_contents = await redis.get("temp_docmuncher_contents")
+        num_pages = len(markdown_text["pages"])
+
+        # Update expected contents since task has finished
+        await redis.set(
+            "temp_docmuncher_contents",
+            min(
```
lickem22, line 466: Shouldn't this be max?

lickem22 (approved):
Hi @poornimaramesh  thank you so much for this.
There are a few things to fix :
- The `temp_docmuncher_contents` is not set per workspace.
- I have this weird bug where where it seems like the `temp_docmuncher_contents` doesn't get updated. So when you add a failed doc (adding the same document twice) and you try to add a valid document it doesn't allow you.
We can discuss that in a huddle tomorrow. But I can also  fix if you don't have the capacity.

core_backend/app/docmuncher/routers.py
poornimaramesh: fixed it -- it's because the reset in `dependencies` wasn't computing the number of pages correctly

poornimaramesh:
@lickem22 added a few more changes as we discussed:
1. A default page quota (switch this to 10 in your local repo for testing)
2. Now we do 2 checks: (a) whether the upload directly exceeds this quota, (b) whether the number of cards that could potentially be created from this exceed the content quota
3. The error message from (a) and (b) are both written in terms of pages to enable better UX.

Repeat the same tests as above to check if things work as expected

poornimaramesh:
> Hi @poornimaramesh thank you so much for this. There are a few things to fix :
> 
> * The `temp_docmuncher_contents` is not set per workspace.
> * I have this weird bug where where it seems like the `temp_docmuncher_contents` doesn't get updated. So when you add a failed doc (adding the same document twice) and you try to add a valid document it doesn't allow you.
>   We can discuss that in a huddle tomorrow. But I can also  fix if you don't have the capacity.

This bug was because I was using the markdown text to determine number of pages / contents to decrease from the temp variable. But in the tags case, obviously no markdown text is created. Now I use the input file content in `process_pdf_file` to determine the number of pages, so it seems to work

core_backend/app/users/models.py
```diff
@@ -114,6 +114,7 @@ class WorkspaceDB(Base):
         DateTime(timezone=True), nullable=True
     )
     content_quota: Mapped[int | None] = mapped_column(Integer, nullable=True)
+    page_quota: Mapped[int | None] = mapped_column(Integer, nullable=True)
```
lickem22, line 117: I think we should not add it in the DB. Since we keep the page quota as a factor of the contents quota we can just compute instead of saving to DB. I think that would make things easier.

lickem22 (approved):

PR #561, opened 2025-10-07
Adding document language attribute to workspace DB and updating translation process for search

tonyzhao6:
Reviewer:
Estimate: 30 minutes

---

## Description

Current search process incorrectly skips translation if the user query is in English/always translates user query to English without checking the language of the indexed documents. Instead, we should translate the user query to match the language of the indexed documents.

### Changes

1. Added a `doc_language` attribute to `WorkspaceDB`.
2. Updated search endpoint pipeline to translate user query only if needed to match the language of the indexed documents.
3. Updated `litellm` docker version to latest.
4. Update litellm config to also contain `VOYAGE_API_KEY`.

### Future Tasks (optional)

1. Next PR will update the frontend UI to allow users to specify/update the document language for their workspace.

## How has this been tested?

1. All pre-existing tests have been updated to support the new `doc_language` attribute.
2. New tests added for `doc_language` attribute.
5. UI tested by manually changing workspace language to `FRENCH`.

## Checklist

Fill with `x` for completed. 

- [X] I have updated the automated tests
- [X] I have updated the requirements


Makefile
```diff
@@ -86,7 +86,7 @@ setup-llm-proxy:
 	-@docker rm litellm-proxy
 	@docker system prune -f
 	@sleep 2
-	@docker pull ghcr.io/berriai/litellm:main-v1.40.10
+	@docker pull ghcr.io/berriai/litellm:main-v1.77.3-stable
```
tonyzhao6, line 89: This update should allow us to use newer models such as the Voyage AI models.

core_backend/app/auth/routers.py
```diff
@@ -240,6 +240,7 @@ async def authenticate_or_create_google_user(
             api_daily_quota=DEFAULT_API_QUOTA,
             asession=asession,
             content_quota=DEFAULT_CONTENT_QUOTA,
+            doc_language=DEFAULT_DOC_LANGUAGE,
```
tonyzhao6, line 243: When a new user registers via google credentials, their workspace document language will initially be `ENGLISH`. The user can then update this value in the UI after logging in.

core_backend/app/users/routers.py
```diff
@@ -258,7 +260,9 @@ async def create_first_user(
     # 1.
     user.role = UserRoles.ADMIN
     user.workspace_name = user.workspace_name or f"{user.username}'s Workspace"
-    workspace_db_new, _ = await create_workspace(asession=asession, user=user)
+    workspace_db_new, _ = await create_workspace(
+        asession=asession, doc_language=IdentifiedLanguage.ENGLISH, user=user
```
tonyzhao6, line 264: For the very first user, the default document language is `ENGLISH`. the user can then update this in the UI after logging in.

lickem22 (approved):
Thanks @tonyzhao6, Everything works well. 
Just one thing I noticed is when you edit a workspace with the same workspace name, it fails. Do you know why?

tonyzhao6:
> Thanks @tonyzhao6, Everything works well. Just one thing I noticed is when you edit a workspace with the same workspace name, it fails. Do you know why?

do you mean as in this screenshot?

Screenshot 2025-10-08 at 08 24 27

if so, i believe this is the expected behavior; i.e., the user cannot change the workspace name to something that already exist---thus, we display that warning message.

tonyzhao6:
> > > Thanks @tonyzhao6, Everything works well. Just one thing I noticed is when you edit a workspace with the same workspace name, it fails. Do you know why?
> > 
> > 
> > do you mean as in this screenshot?
> > Screenshot 2025-10-08 at 08 24 27
> > if so, i believe this is the expected behavior; i.e., the user cannot change the workspace name to something that already exist---thus, we display that warning message.
> 
> Yes, it made sense when we could only edit workspace name. Now lets say you have workspace1 with language as English. If now within the backend to workspace1 with language as French it fails saying: workspace with name `workspace1` already exists. It shouldn't happen

gotcha---yes, this would be addressed by the next PR with the frontend UI change because right now, there is no way to change the language except via the global config. did you want the backend ability in this PR as well?

tonyzhao6:
> > > > > Thanks @tonyzhao6, Everything works well. Just one thing I noticed is when you edit a workspace with the same workspace name, it fails. Do you know why?
> > > > 
> > > > 
> > > > do you mean as in this screenshot?
> > > > Screenshot 2025-10-08 at 08 24 27
> > > > if so, i believe this is the expected behavior; i.e., the user cannot change the workspace name to something that already exist---thus, we display that warning message.
> > > 
> > > 
> > > Yes, it made sense when we could only edit workspace name. Now lets say you have workspace1 with language as English. If now within the backend to workspace1 with language as French it fails saying: workspace with name `workspace1` already exists. It shouldn't happen
> > 
> > 
> > gotcha---yes, this would be addressed by the next PR with the frontend UI change because right now, there is no way to change the language except via the global config. did you want the backend ability in this PR as well?
> 
> I was able to change the language in the DB with the edit backend tho. Are you sure it's not ?

got it, understand what you mean now! thanks for the catch---latest updates should reflect the correct behavior.

PR #566, opened 2025-11-10
Adding background task capability to chat endpoint

tonyzhao6:
Reviewer: @suzinyou 
Estimate: 30 minutes

---

### Goal

1. Add background task capability to question answer chat endpoint for Turn.io.

### Changes

1. Changes are documented in comments below.

## How has this been tested?

1. In the UI, I created content and tested the chat feature. This should go thru the regular async route and return a response (it did).
2. In Swagger docs (using API key generated from UI), I tested the `/chat` endpoint with `turnio_api_key` and `wa_id` set to `null` and verified that we get the usual async response back.
3. In Swagger docs, with `turnio_api_key` and `wa_id` specified to some dummy values, I verified that I get an immediate response back, the background executes the search and LLM response generation (can see this in terminal), and then **errors out when making the call to Turn since I can't test this without deployment and an actual number.**
4. Will have to deploy to production and test with real Turn API key and WA ID.

## Checklist

Fill with `x` for completed. 

- [X] I have updated the automated tests

### DIFF
--- a/core_backend/app/question_answer/routers.py
+++ b/core_backend/app/question_answer/routers.py
@@ -155,8 +155,16 @@ async def _chat(
 
         # 3.
         if user_query.turnio_api_key and user_query.wa_id:
+            if isinstance(response, JSONResponse):
+                raw = response.body
+                if isinstance(raw, memoryview):
+                    raw = raw.tobytes()
+                data = json.loads(raw.decode("utf-8"))
+                text_body = data["error_message"]
+            else:
+                text_body = response.llm_response
             payload = TurnTextMessage.model_validate(
-                {"to": user_query.wa_id, "text": {"body": response.llm_response}}
+                {"to": user_query.wa_id, "text": {"body": text_body}}
             )
             payload_dict = payload.model_dump()
             response_json = await send_turn_text_message(
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -91,6 +91,10 @@
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
     {
+      "path": "detect_secrets.filters.common.is_baseline_file",
+      "filename": ".secrets.baseline"
+    },
+    {
       "path": "detect_secrets.filters.common.is_ignored_due_to_verification_policies",
       "min_level": 2
     },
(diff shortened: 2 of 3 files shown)


core_backend/app/config.py
```diff
@@ -120,3 +120,9 @@
 
 # Language
 DEFAULT_DOC_LANGUAGE = os.environ.get("DEFAULT_DOC_LANGUAGE", "ENGLISH")
+
+# HTTPX
+HTTPX_TIMEOUT: float = 10.0
+
+# Turn.io
+TURNIO_API_BASE: str = os.environ.get("TURNIO_API_BASE", "https://whatsapp.turn.io/v1/")
```
tonyzhao6: For AAQ, we only need `TURNIO_API_BASE`, which is fixed. The `TURNIO_API_KEY` will be provided in each `QueryBase` object instead.

core_backend/app/question_answer/schemas.py
```diff
@@ -37,6 +37,8 @@ class QueryBase(BaseModel):
     )
     query_text: str = Field(..., examples=["What is AAQ?"])
     session_id: int | None = Field(default=None)
+    turnio_api_key: Optional[str] = None
```
tonyzhao6: `turnio_api_key` and `wa_id` are now part of `QueryBase` as optionals. When **both** are specified, then the `chat` endpoint will run in the background. Otherwise, the `chat` endpoint will run async as usual.

core_backend/app/question_answer/routers.py
```diff
@@ -91,6 +95,93 @@
...
+    Returns
+    -------
+    QueryResponse | JSONResponse | dict[str, Any]
+        The query response object, an appropriate JSON response, or a dictionary
+        (for WhatsApp responses).
+    """
+
+    try:
+        # 1.
+        user_query = await init_user_query_and_chat_histories(
```
tonyzhao6: Initializing user query and chat histories is the same for both async and background task logic.

core_backend/app/question_answer/routers.py
```diff
@@ -91,6 +95,93 @@
...
+                turnio_api_key=user_query.turnio_api_key,
+                whatsapp_id=user_query.wa_id,
+            )
+            payload_dict.update(response_json)
+            langfuse_context.update_current_trace(name="chat")
+            return payload_dict
+
+        # 4.
+        return response
+    except LLMCallException:
```
tonyzhao6: Both logic has the same except clause to catch any errors as a `JSONResponse`.

suzinyou (commented):
Code looks good! Will look through the tests and run them before approving

suzinyou (approved):

